PR #1, opened 2016-12-14
Adds support for optimistic concurrency

SzymonPobiega:
BTW, I think we should implement the pessimistic locking eventually. But this one is good for start. I haven't run the MySQL version but it should be fine as it uses plain old `int` column and does the incrementing when saving.

The version is stored in the context bag which makes it hard to use custom finders. Basically a custom finder needs to set this the version value for the returned saga. It is even more tricky since custom finders don't have access to writable context and need to downcast.

An alternative would be to add a field in `IContainSagaData` derived class but it feels wrong to me. The version should be handled internally and not exposed on the saga data.

### DIFF
--- a/ScriptBuilder/Saga/Writers/MsSqlServerSagaScriptWriter.cs
+++ b/ScriptBuilder/Saga/Writers/MsSqlServerSagaScriptWriter.cs
@@ -159,7 +159,8 @@ set @createTable = '
         OriginalMessageId nvarchar(255),
         Data nvarchar(max) not null,
         PersistenceVersion varchar(23) not null,
-        SagaTypeVersion varchar(23) not null
+        SagaTypeVersion varchar(23) not null,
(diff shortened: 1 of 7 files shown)


SimonCropp:
> The version is stored in the context bag which makes it hard to use custom finders. Basically a custom finder needs to set this the version value for the returned saga. It is even more tricky since custom finders don't have access to writable context and need to downcast.

Perhaps we can handle this with some helper/extension methods?

SimonCropp:
verified it works on mysql

SimonCropp:
happy with this to be merged as is for now and worry about finder stuff later

SzymonPobiega:
When I think about it now, using `SynchronizedStorageSession` might be the best option. It is accessible and we can add an extension method to it. I've checked the acceptance tests and they are all green.

PR #7, opened 2017-01-30
Switching to sequential access for outbox, saga and timeout persister

danielmarbach:
Haven't had a chance to thoroughly test it. We applied the same for Sql Server [in this PR
](https://github.com/Particular/NServiceBus.SqlServer/pull/211)

More information see

https://msdn.microsoft.com/en-us/library/hh556234.aspx

and

https://stackify.com/improved-performance-using-json-streaming/

### DIFF
--- a/SqlPersistence/Outbox/OutboxPersister.cs
+++ b/SqlPersistence/Outbox/OutboxPersister.cs
@@ -65,22 +65,23 @@ class OutboxPersister : IOutboxStorage
                     command.CommandText = outboxCommands.Get;
                     command.Transaction = transaction;
                     command.AddParameter("MessageId", messageId);
-                    using (var dataReader = await command.ExecuteReaderAsync(CommandBehavior.SingleRow))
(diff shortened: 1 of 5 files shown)


danielmarbach:
Btw technically if we can create a dedicated DbCommand object it would also be possible to stream data into the server using this approach

               using (StreamReader stream = ...) {

                  // Add a parameter which uses the StreamReader we just opened
                  // Size is set to -1 to indicate "MAX"
                  cmd.Parameters.Add("@textdata", SqlDbType.NVarChar, -1).Value = stream ;

                  // Send the data to the server asynchronously
                  await cmd.ExecuteNonQueryAsync();
               }

more information can be found in the streaming data to SQL Server section of the document linked above.

SimonCropp:
@danielmarbach 

> Btw technically if we can create a dedicated DbCommand object it would also be possible to stream data into the server using this approach

I had a go at this previously but could not get it working. perhaps a pairing session?

PR #10, opened 2017-02-08
[WIP] Switched to using a nonclustered index on message Id

SzymonPobiega:


SimonCropp:
@SzymonPobiega 

 * isnt `unique` redundant since it is already a primary key?

So shouldnt it look like this?


```
        MessageId nvarchar(200) not null primary key nonclustered,
        Dispatched bit not null default 0,
        DispatchedAt datetime,
        PersistenceVersion varchar(23) not null,
        Operations nvarchar(max) not null
```

SzymonPobiega:
@SimonCropp true. At first I wanted to just add a unique index without PK but then I messed up and added both ;-) Still I don't know what is the actual difference between a non-clustered PK and a non-clustered unique index. Need to google that up.

SimonCropp:
@SzymonPobiega can u rebase and re-run the approval tests.

How dow we transition existing users? a docs PR for an upgrade guide for existing users, ie a script they can run?  or do it in the installer?

SzymonPobiega:
@SimonCropp How about adding info to the release notes? The creation scripts are generated anyway so users can use their favorite SQL diff tool to generate the transition script. I would be afraid to run these kinds of scripts in the installers.

SimonCropp:
Yep happy with release notes

SimonCropp:
I will work on a draft upgrade guide on Monday.

SimonCropp:
replaced by https://github.com/Particular/NServiceBus.Persistence.Sql/pull/33

PR #14, opened 2017-02-16
use SequentialGuid for timeout id

### LINKED ISSUE #13: Guid is used as a timeout ID
The effect is index fragmentation and low performance of index operations.

A better solution is to use CombGuids.

SimonCropp:
fixes #13

### DIFF
--- a/SqlPersistence/Timeout/SequentialGuid.cs
+++ b/SqlPersistence/Timeout/SequentialGuid.cs
@@ -1,0 +1,33 @@
+namespace NServiceBus.Persistence.Sql
+{
+    using System;
+    using System.Threading;
+
+    // From https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore/ValueGeneration/SequentialGuidValueGenerator.cs
+    static class SequentialGuid
+    {
+        static long counter = DateTime.UtcNow.Ticks;
+
+        internal static Guid Next()
+        {
+            var guidBytes = Guid.NewGuid().ToByteArray();
+            var counterBytes = BitConverter.GetBytes(Interlocked.Increment(ref counter));
+
+            if (!BitConverter.IsLittleEndian)
+            {
+                Array.Reverse(counterBytes);
+            }
+
+            guidBytes[08] = counterBytes[1];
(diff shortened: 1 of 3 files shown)


SqlPersistence/Timeout/SequentialGuid.cs
```diff
@@ -0,0 +1,33 @@
+namespace NServiceBus.Persistence.Sql
+{
+    using System;
+    using System.Threading;
+
+    // From https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore/ValueGeneration/SequentialGuidValueGenerator.cs
+    static class SequentialGuid
+    {
+        static long counter = DateTime.UtcNow.Ticks;
```
SzymonPobiega: Is this algorithm safe? I guess the change that of conflict is small when using ticks but maybe we should include some machine identification also (MAC?). Like they propose here: http://stackoverflow.com/questions/1752004/sequential-guid-generator

SimonCropp: It is what EF uses ;)

SzymonPobiega: :trollface: Well, we can't just use the `Win32` call because of portability. Should we call `YAGNI` on it and revisit it if we actually experience conflicts?

SimonCropp: according to this https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore/ValueGeneration/SequentialGuidValueGenerator.cs#L11

>  Generates sequential <see cref="Guid" /> values using the same algorithm as NEWSEQUENTIALID()

so i think it is safe

cremor:
Is there a specific reason why the EF implementation was used instead of the one in [CombGuid.cs](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/IdGeneration/CombGuid.cs) in NServiceBus.Core?

SimonCropp:
@cremor i wanted to use the one closest to the [sqlserver NEWSEQUENTIALID method](https://msdn.microsoft.com/en-us/library/ms189786.aspx) implementation

PR #30, opened 2017-02-24
When dispatched empty Outbox operations to free up storage

DavidBoike:
When the Outbox document is marked as dispatched, the outgoing transport operations are no longer needed. This update will clear the transport operations at that point, saving space in the database, and making it more efficient to remove Outbox records that have expired.


SimonCropp:
@DavidBoike thanks. Not possible to add a test?

@SzymonPobiega given u wrote most of that that can u review this

DavidBoike:
I'm not 100% sure, I ran the unit tests locally and then pushed it up to see if it would break any acceptance tests. Kind of expected it to, but it didn't, then I got sidetracked. (Are those currently run on TeamCity?)

Best option would probably be store some operations, mark as dispatched, then retrieve and ensure you get back empty set. I can add that Monday.

SimonCropp:
Yep acceptance test are running on TC

Issue #21, opened 2017-02-20
Add ability to promote SQL scripts to source directory


DavidBoike:
https://docs.particular.net/nservicebus/sql-persistence/#script-creation-promotion

SimonCropp:
@weralabaj @SzymonPobiega this is a XS that would be a good one for someone to get some knowledge of MSBuild. any takers for a PR?

weralabaj:
I'm interested, but can start tomorrow, if nobody else beats me up to this, OK?

PR #34, opened 2017-02-27
Custom script location

weralabaj:
https://github.com/Particular/NServiceBus.Persistence.Sql/issues/21


SimonCropp:
sorry i didnt communicate this very well. we want to "replicate the scripts to anther location" not "save them to a custom location". the reason is we still want them to be added to the build assets https://github.com/Particular/NServiceBus.Persistence.Sql/blob/develop/packaging/nuget/NServiceBus.Persistence.Sql.MsBuild.targets#L26

weralabaj:
@SimonCropp better now?

src/ScriptBuilderTask/ScriptBuilderTask.cs
```diff
@@ -74,13 +80,22 @@ void Inner()
             }
             foreach (var variant in SqlVariantReader.Read(moduleDefinition))
             {
-                Write(moduleDefinition, variant);
+                Write(moduleDefinition, variant, scriptPath);
+            }
+
+            var customPath = OutputPathReader.Read(moduleDefinition);
+            if (string.IsNullOrEmpty(customPath) == false)
```
SzymonPobiega: Invert the `If` to reduce nesting and `return` here is no custom path?

SzymonPobiega:
Is there any option to have an acceptance test for that that verifies:
 * That the scripts are actually replicated
 * That ProjectPath and SolutionPath variables are substituted in the custom path?

src/ScriptBuilderTask/ScriptBuilderTask.cs
```diff
@@ -83,14 +95,40 @@ void Inner()
...
+        }
+
+        void DuplicateDirectory(string source, string destination)
+        {
+            foreach (var dirPath in Directory.GetDirectories(source, "*", SearchOption.AllDirectories))
+            {
+                Directory.CreateDirectory(dirPath.Replace(source, destination));
+            }
+
+            foreach (var newPath in Directory.GetFiles(source, "*.*", SearchOption.AllDirectories))
```
weralabaj: Is that to avoid the cost of generating the content?

src/ScriptBuilder/ScriptPromotionPathReader.cs
```diff
@@ -0,0 +1,31 @@
...
+            .FirstOrDefault(x => x.AttributeType.FullName == "NServiceBus.Persistence.Sql.SqlPersistenceSettingsAttribute");
+        if (customAttribute == null)
+        {
+            target = null;
+            return false;
+        }
+
+        var arguments = customAttribute.ConstructorArguments;
+        target = (string) arguments[2].Value;
+        if (target == null)
```
weralabaj: is it even possible, since we provide default values?

SimonCropp: i changed the default to null https://github.com/Particular/NServiceBus.Persistence.Sql/blob/develop/src/SqlPersistence/SqlPersistenceSettingsAttribute.cs#L15

src/ScriptBuilderTask/ScriptBuilderTask.cs
SimonCropp: yes

PR #35, opened 2017-02-28
WIP: Custom finder helper

SimonCropp:


SimonCropp:
@Particular/sqlpersistence-maintainers before i go too much further this PR enabled a finder API if

```
class OrderSagaFinder :
    IFindSagas<OrderSagaData>.Using<PaymentTransactionCompleted>
{
    public Task<OrderSagaData> FindBy(PaymentTransactionCompleted message, SynchronizedStorageSession session, ReadOnlyContextBag context)
    {
        return session.GetSagaData<OrderSagaData>(
            readOnlyContextBag: context,
            whereClause: "JSON_VALUE(Data,'$.PaymentTransactionId') = @propertyValue",
            addParameters: command =>
            {
                var parameter = command.CreateParameter();
                parameter.ParameterName = "propertyValue";
                parameter.Value = message.PaymentTransactionId;
                command.Parameters.Add(parameter);
            });
    }
}
```

that do u think?

weralabaj:
Stupid question, but from the API perspective, the only things that changed are method name and that we don't need to pass endpoint name in the FindBy method, correct?

From the implementation perspective I think it's much cleaner now. Not sure what other goals were there?

SimonCropp:
@weralabaj sorry. here is some more context https://docs.particular.net/samples/saga/sql-sagafinder/#sql-persistence-helpers

weralabaj:
Ah now it makes sense, I looked at the api in the sample and it looked so similar :P Sorry for stupid question, API LGTM!

andreasohlund:
@SimonCropp looked at the code and I retract my concerns since the `SynchronizedStorageSession` is just a marker interface we don't leak irrelevant things into the finders.

That said (and YAGNI very likely applies here) if we ever add things to the SSS that isn't relevant for saga finders it might confuse users.

In short: I had a thought that we should go with a custom "SagaStorageSession" 
abstraction to be in full controll over the API  but this is close enough

src/SqlPersistence/SqlPersistenceStorageSessionExtensions.cs
```diff
@@ -4,6 +4,9 @@
 
 namespace NServiceBus
 {
+    using System.Data.Common;
+    using System.Threading.Tasks;
+    using Extensibility;
 
     /// <summary>
     /// Shared session extensions for NHibernate persistence.
```
weralabaj: copy paste?

SimonCropp: https://github.com/Particular/NServiceBus.Persistence.Sql/commit/63ae3486c799c9219fff386d103bb6c37046f4d8

and will be included when i do a cleaner PR

src/SqlPersistence/SqlPersistenceStorageSessionExtensions.cs
```diff
@@ -15,12 +18,36 @@ public static class SqlPersistenceStorageSessionExtensions
         /// </summary>
         public static ISqlStorageSession SqlPersistenceSession(this SynchronizedStorageSession session)
         {
+            return GetSqlStorageSession(session);
+        }
+
+        internal static StorageSession GetSqlStorageSession(this SynchronizedStorageSession session)
```
weralabaj: Inconsistent naming: here `SqlPersistenceSession`, NHIbernate `Session()`, Raven `RavenSession()`

> "Shared session has not been configured for SqlPersistence."

Is that something that users need to configure? How? Exception message doesn't seem to be very helpful.

src/MsSqlAcceptanceTests/When_custom_finder_returns_existing_saga.cs
```diff
@@ -0,0 +1,114 @@
...
+            public Context Context { get; set; }
+
+            public Task<TestSaga.SagaData> FindBy(SomeOtherMessage message, SynchronizedStorageSession session, ReadOnlyContextBag context)
+            {
+                Context.FinderUsed = true;
+
+                return session.GetSagaData<TestSaga.SagaData>(
+                    readOnlyContextBag: context,
+                    whereClause: "JSON_VALUE(Data,'$.Property') = @propertyValue",
+                    addParameters: command =>
```
weralabaj: > addParameters

I don't get this parameter
- why do we allow to pass `Action<DbCommand>`s here?
- is `whereClause` not sufficient to pass the desired query?
- why is that called "addParameters" if I can pass commands with text, not only parameters?

SimonCropp: @weralabaj because we own the start of the command text. they own the where and what parameters are passed through

weralabaj: So wouldn't it make sense to only allow passing parameters i.e. DbParameter rather than DBCommand?

SimonCropp: then how would they call https://msdn.microsoft.com/en-us/library/system.data.common.dbcommand.createparameter(v=vs.110).aspx  ?

weralabaj: they shouldn't, IMO we should call it, maybe let them just pass key-values, or provide some custom type for this?
either way, since I can't set other elements of the command, it shouldn't be exposed there, it's a small WTF at first look

SimonCropp: @weralabaj how you add parameters is dependent on type. so a `dictionary<string,object>` wont work

weralabaj: I still hate it, but I guess there's how it has to be then

SimonCropp: lol. hate. yeah i dont like it either. but could not come up with anything better

src/SqlPersistence/Saga/RuntimeSagaInfo.cs
```diff
@@ -36,7 +38,7 @@ class RuntimeSagaInfo
         Type sagaType,
         NewtonSerializer jsonSerializer,
         Func<TextReader, JsonReader> readerCreator,
-        Func<TextWriter, JsonWriter> writerCreator)
+        Func<TextWriter, JsonWriter> writerCreator, string tablePrefix)
```
SzymonPobiega: nitpicking: newline before `tablePrefix`?

src/MsSqlAcceptanceTests/When_custom_finder_returns_existing_saga.cs
SzymonPobiega: `Func<Func<DbParamete>, IEnumerable<DbParameter>>` ? A func that takes a parameter factory and returns a list of parameter object. I agree with @weralabaj that exposing `DbCommand` is confusing.

src/SqlPersistence/SqlPersistenceStorageSessionExtensions.cs
SzymonPobiega: I've copied it over from NHIbernate. I agree it is not particularly helpful. User's don't configure it. You would get it if you would reference two persistence mechanisms and used, say, NH for the endpoint but in a handler you would use Sql persister extension method. Pretty unlikely.

SimonCropp:
replaced by https://github.com/Particular/NServiceBus.Persistence.Sql/pull/42

PR #43, opened 2017-03-08
explicit Schema support

SimonCropp:


src/ScriptBuilder.Tests/Saga/SagaDefinitionValidatorTest.WithInvalidSuffixLeft.approved.txt
```diff
@@ -0,0 +1 @@
+﻿The Saga 'saga1' has a tableSuffix that contains a '['. Delimiters are automatically and are not required in configuration.
```
SzymonPobiega: automatically added?

src/ScriptBuilder/Saga/SagaDefinitionValidator.cs
```diff
@@ -14,9 +14,24 @@ internal static void ValidateSagaDefinition(string correlationProperty, string s
...
+            {
+                throw new ErrorsException($"The Saga '{sagaName}' has an empty string defined for to TableSuffix.");
+            }
+            if (tableSuffix.Contains("]"))
+            {
+                throw new ErrorsException($"The Saga '{sagaName}' has a tableSuffix that contains a ']'. Delimiters are automatically added and are not required in configuration.");
+            }
+            if (tableSuffix.Contains("["))
+            {
+                throw new ErrorsException($"The Saga '{sagaName}' has a tableSuffix that contains a '['. Delimiters are automatically and are not required in configuration.");
```
SzymonPobiega: automatically added

src/ScriptBuilder/Saga/SagaDefinitionValidator.cs
```diff
@@ -14,9 +14,24 @@ internal static void ValidateSagaDefinition(string correlationProperty, string s
...
+            }
+            if (tableSuffix.Contains("`"))
+            {
+                throw new ErrorsException($"The Saga '{sagaName}' has a tableSuffix that contains a '`'. Delimiters are automatically and are not required in configuration.");
```
SzymonPobiega: automatically added.

SzymonPobiega: Also, I think it is better to say that the names are *automatically quoted* instead of saying we add delimiter. I would be confused as I would tread the `.` as a delimiter in the fully-qualified name, not the `[]`s

src/SqlPersistence.Tests/APIApprovals.Approve.approved.txt
```diff
@@ -112,6 +112,7 @@ namespace NServiceBus
         public static void ConnectionBuilder(this NServiceBus.PersistenceExtensions<NServiceBus.Persistence.Sql.SqlPersistence> configuration, System.Func<System.Data.Common.DbConnection> connectionBuilder) { }
         public static void DisableInstaller(this NServiceBus.PersistenceExtensions<NServiceBus.Persistence.Sql.SqlPersistence> configuration) { }
         public static NServiceBus.Persistence.Sql.SagaSettings SagaSettings(this NServiceBus.PersistenceExtensions<NServiceBus.Persistence.Sql.SqlPersistence> configuration) { }
+        public static void Schema(this NServiceBus.PersistenceExtensions<NServiceBus.Persistence.Sql.SqlPersistence> configuration, string schema) { }
```
SzymonPobiega: Is there a way to introduce variant-specific config API? Asking because in Oracle schema is something totally different than in SQL Server.

SzymonPobiega:
@SimonCropp I am still not sure why we arbitrarily forbid usage of `[` and `]`. While I do believe using them in the name would not be a good practice for sure, I would feel better if we had proper arguments why we don't allow them (the current exception text does not say anything about the reason).

src/ScriptBuilder/Saga/SagaDefinitionValidator.cs
SimonCropp: fixed

src/SqlPersistence.Tests/APIApprovals.Approve.approved.txt
SimonCropp: would we not have the same config for oracle?

SimonCropp:
> I am still not sure why we arbitrarily forbid usage of [ and ]. While I do believe using them in the name would not be a good practice for sure, I would feel better if we had proper arguments why we don't allow them (the current exception text does not say anything about the reason).

Well the simple reason is to trivially to catch this `persistence.TablePrefix("[My Schema].");` case as part of the update https://docs.particular.net/nservicebus/sql-persistence/#installation-table-prefix-database-schema

but wouldnt you say encouraging people to exclude those characters is helping people make good decisions?

src/ScriptBuilder.Tests/Saga/SagaDefinitionValidatorTest.WithInvalidSuffixLeft.approved.txt
SimonCropp: fixed

src/SqlPersistence.Tests/APIApprovals.Approve.approved.txt
SzymonPobiega: AFAIK Schema in Oracle is more-or-less equivalent of Catalog in SQL Server: https://docs.microsoft.com/en-us/sql/ssma/oracle/mapping-oracle-schemas-to-sql-server-schemas-oracletosql

SimonCropp: so it is not a valid config for oracle?

SzymonPobiega:
@SimonCropp well, I get the upgrade argument. I am a bit worried that this is not aligned with what SQL transport expects (it can tolerate anything).

src/SqlPersistence.Tests/APIApprovals.Approve.approved.txt
SzymonPobiega: @SimonCropp I believe it would be valid but would mean something different. @gbiellem @MarcinHoppe you are our Oracle expects. What do you think?

SimonCropp:
@SzymonPobiega yeah i am not sure "because another lib enables bad practices that this one should as well"?

SimonCropp:
@SzymonPobiega how about we keep the character check in place for 2.0, to catch the people possibly using the current schema syntax. and we can always relax that restriction in a future release?

SzymonPobiega:
@SimonCropp I ❤️ it.

Issue #40, opened 2017-03-01
Is the SqlSagaAttribute really needed to detect the correlation property?


danielmarbach:
I was looking briefly into the saga definition reader and wrote the following simple mapping:

        protected override void ConfigureHowToFindSaga(SagaPropertyMapper<SagaData> mapper)
        {
            mapper.ConfigureMapping<object>(o => null).ToSaga(s => s.Correlation);
            mapper.ConfigureMapping<string>(o => null).ToSaga(s => s.Correlation);
        }

Which produced the following IL method body

```
items = {Mono.Cecil.Cil.Instruction[]} {Mono.Cecil.Cil.Instruction[122]}
 [0] = {Mono.Cecil.Cil.Instruction} {IL_0000: nop}
 [1] = {Mono.Cecil.Cil.Instruction} {IL_0001: ldarg.1}
 [2] = {Mono.Cecil.Cil.Instruction} {IL_0002: ldtoken System.Object}
 [3] = {Mono.Cecil.Cil.Instruction} {IL_0007: call System.Type System.Type::GetTypeFromHandle(System.RuntimeTypeHandle)}
 [4] = {Mono.Cecil.Cil.Instruction} {IL_000c: ldstr "o"}
 [5] = {Mono.Cecil.Cil.Instruction} {IL_0011: call System.Linq.Expressions.ParameterExpression System.Linq.Expressions.Expression::Parameter(System.Type,System.String)}
 [6] = {Mono.Cecil.Cil.Instruction} {IL_0016: stloc.0}
 [7] = {Mono.Cecil.Cil.Instruction} {IL_0017: ldnull}
 [8] = {Mono.Cecil.Cil.Instruction} {IL_0018: ldtoken System.Object}
 [9] = {Mono.Cecil.Cil.Instruction} {IL_001d: call System.Type System.Type::GetTypeFromHandle(System.RuntimeTypeHandle)}
 [10] = {Mono.Cecil.Cil.Instruction} {IL_0022: call System.Linq.Expressions.ConstantExpression System.Linq.Expressions.Expression::Constant(System.Object,System.Type)}
 [11] = {Mono.Cecil.Cil.Instruction} {IL_0027: ldc.i4.1}
 [12] = {Mono.Cecil.Cil.Instruction} {IL_0028: newarr System.Linq.Expressions.ParameterExpression}
 [13] = {Mono.Cecil.Cil.Instruction} {IL_002d: dup}
[...]

PR #46, opened 2017-03-09
[SPIKE] Infer correlation property from ConfigureMapping IL if possible

DavidBoike:
I had this half done locally before #40 was closed, and I think I succeeded.

It's based upon being very conservative and assuming the user will stay within well-established conventions for the ConfigureHowToFindSaga method, and bails out (deferring to the attribute) if anything is out of place.

This is the flow:

* Attempt to access the attribute first, it is always overriding
* Bail out if the saga class does not directly inherit NServiceBus.Saga<T> so that nothing funky can happen in an intermediate base class
* Within the IL of the ConfigureHowToFindSaga method, we should be OK if there is no branching logic and we don't call into any method or delegate that isn't specifically allowed
  * Bail out if any branching OpCode is encountered
  * Inspect every Call/Calli/Callvirt OpCode, bailing out if any unknown method is called
    * Attempt to optimize perf by performing string equality operations before StartsWith operations for generic types
  * Inspect Ldtoken OpCodes where the declaring type is the saga data type, as these identify the `saga.Correlation` portion of the `.ToSaga(saga => saga.Correlation)` calls.
    * Ensure that all these calls within the method point to the same property, otherwise bail out

I have tests for:

* Mapping a single message type
* Mapping two message types
* Mapping with an int CorrelationId to cover boxing instructions
* Failure cases for:
  * Saga with a non-abstract intermediate base class
  * Method is called within the .ToSaga() expression
  * SagaPropertyMapper is passed to another method
  * Two messages mapped with conflicting correlation ids
  * If/then branching logic used in mapping method
  * ConfigureMapping called within an Action delegate
  * Looping used in mapping method

@SimonCropp @danielmarbach

### DIFF
--- a/src/ScriptBuilder.Tests/Saga/InferCorrelationIdTests.DualMapping.approved.txt
+++ b/src/ScriptBuilder.Tests/Saga/InferCorrelationIdTests.DualMapping.approved.txt
@@ -1,0 +1,9 @@
+﻿{
+  "TableSuffix": "DualMappingSaga",
+  "CorrelationProperty": {
+    "Type": "String",
+    "Name": "Correlation"
+  },
+  "TransitionalCorrelationProperty": null,
+  "Name": "InferCorrelationIdTests/DualMappingSaga"
+}
--- a/src/ScriptBuilder.Tests/Saga/InferCorrelationIdTests.SingleMapping.approved.txt
+++ b/src/ScriptBuilder.Tests/Saga/InferCorrelationIdTests.SingleMapping.approved.txt
@@ -1,0 +1,9 @@
+﻿{
+  "TableSuffix": "SingleMappingSaga",
+  "CorrelationProperty": {
+    "Type": "String",
+    "Name": "Correlation"
+  },
+  "TransitionalCorrelationProperty": null,
+  "Name": "InferCorrelationIdTests/SingleMappingSaga"
+}
--- a/src/ScriptBuilder.Tests/Saga/InferCorrelationIdTests.SingleMappingValueType.approved.txt
+++ b/src/ScriptBuilder.Tests/Saga/InferCorrelationIdTests.SingleMappingValueType.approved.txt
@@ -1,0 +1,9 @@
+﻿{
+  "TableSuffix": "SingleMappingValueTypeSaga",
+  "CorrelationProperty": {
+    "Type": "Int",
+    "Name": "Correlation"
+  },
+  "TransitionalCorrelationProperty": null,
+  "Name": "InferCorrelationIdTests/SingleMappingValueTypeSaga"
+}
--- a/src/ScriptBuilder.Tests/Saga/InferCorrelationIdTests.cs
+++ b/src/ScriptBuilder.Tests/Saga/InferCorrelationIdTests.cs
@@ -42,6 +42,28 @@ public class InferCorrelationIdTests
     }
 
     [Test]
+    public void SingleMappingValueType()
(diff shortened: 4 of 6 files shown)


SimonCropp:
unless this can detect the edge cases and tell them at compile time they have done something wrong it is not worth the complexity.

also it is a significant amount of fragile code that still results in a worse API then inheriting from `SqlSaga` with an attribute.

SimonCropp:
also the mapping API in the core is clearly flawed. 

 * having to call tosaga multiple times to the same property
 * enables bad practice like branching an base class logic

I would prefer this effort be spent on a PR to the core. so the benefits can be realized for all persistences

SimonCropp:
this also makes the chance of failing at runtime, the first tima a saga is executed, much more common

SimonCropp:
made redundant by https://github.com/Particular/NServiceBus.Persistence.Sql/issues/47

SzymonPobiega:
@SimonCropp  why? As I mentioned in #47 it would be nice if SQL-specific features were treated as enhancements. In this case the script generator should try to understand the expression in the most basic form:

```
mapper.ConfigureMapping<MyMessage>(m => m.SomeId).ToSaga(s => s.SomeId);
```

it it can, all good and it can generate scripts. If not, it should throw an exception saying it can't figure out the property based on the logic and it needs guidance in form of an attribute. Users who don't want (for whatever reason) the attribute can still disable the script generation or (ideally ❤️ ) be allowed to provide equivalent hint outside of the saga class scope.

danielmarbach:
@SzymonPobiega I agree and see my comments here https://github.com/Particular/NServiceBus.Persistence.Sql/issues/40#issuecomment-284321358

SimonCropp:
yes this can be detected

    mapper.ConfigureMapping<MyMessage>(m => m.SomeId).ToSaga(s => s.SomeId);

but put branching logic inside the `m => m.SomeId` and then we cannot.

you are guessing about runtime code in a way that is not reliable. then when we cant detect (in a case that will confuse them as to why) we blow up and tell them to use the attribute approach.

SimonCropp:
basically if we end up suggesting them to use the attribute approach we may as well just get them to do that all the time.

It is a better api, that requires less code and they dont need re declare the same correlation property for every message they want map

i really dont know why you are all pushing so stringly for an API that is clearly worse. 

The argument for "it is easier to migrate to for users of another persister" is fairly weak. people need to change a few lines of code per saga, in a repetitive way, that will result in cleaner code for them. and this effort is dwarfed by the effort that will be required in migrating production data.

danielmarbach:
Pretty sure the comments I made here https://github.com/Particular/NServiceBus.Persistence.Sql/issues/40#issuecomment-284321358 100% agree for ConfigureMapping expression as well. 

You made several times the argument that the new API is simpler and clearer. You also said we should invest in a better core API. So why don't we do that then instead? So arguably this would mean:

1. Sql Persistence uses core saga API and makes assumptions
1. Core changes to better API
1. Sql Persistences bumps core dependency and uses new better API from what we've learned
1. Other persisters can follow

SimonCropp:
@danielmarbach 

> You also said we should invest in a better core API. So why don't we do that then instead? So arguably this would mean:

yes this is a good idea.

SimonCropp:
remember that none of the proposals here get rid of the need for the attribute. we still need it to avoid needing to load all reference assemblies into memory when checking a classes hierarchy

PR #51, opened 2017-03-10
perform cleanup in batch

### LINKED ISSUE #8: Outbox cleanup query uses unbounded DELETE
Currently the Outbox cleaner deletes all the dispatched Outbox records using a single query, regardless of how many records there are. For endpoints with heavy traffic (a lot of new records inserted in a short period of time) this will lead to excessive locking (page locks) and finally to deadlocks and endpoint shutdowns.

Removing many records in one transaction is also quite a heavy operation for the transaction log.

The suggested solution is to use a loop with batched deletes (each in a separate transaction), very much like here:

https://github.com/Particular/NServiceBus.SqlServer/blob/develop/src/NServiceBus.SqlServer/Receiving/ExpiredMessagesPurger.cs#L35-L41

SimonCropp:
fixes #8

### DIFF
--- a/src/SqlPersistence/Outbox/OutboxCommandBuilder.cs
+++ b/src/SqlPersistence/Outbox/OutboxCommandBuilder.cs
@@ -68,12 +68,12 @@ where MessageId = @MessageId";
                 case SqlVariant.MsSqlServer:
                     return $@"
 delete top (@BatchSize) from {tableName}
-where Dispatched = 1
+where Dispatched = 'true'
     and DispatchedAt < @Date";
                 case SqlVariant.MySql:
                     return $@"
 delete from {tableName}
-where Dispatched = 1
+where Dispatched = true
     and DispatchedAt < @Date
 limit @BatchSize";
                 default:
--- a/src/SqlPersistence/Outbox/OutboxCommandBuilder.cs
+++ b/src/SqlPersistence/Outbox/OutboxCommandBuilder.cs
@@ -39,7 +39,12 @@ values
 )";
 
             var cleanupCommand = $@"
-delete from {tableName} where Dispatched = true And DispatchedAt < @Date";
+delete from {tableName}
+where MessageId
+    in (select top @BatchSize MessageId
+        from {tableName}
+        where Dispatched = true
+            and DispatchedAt < @Date)";
 
             var getCommandText = $@"
 select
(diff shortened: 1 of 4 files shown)


SzymonPobiega:
Ready for review?

SimonCropp:
Let me add a test first

SimonCropp:
@SzymonPobiega ready for a review

SimonCropp:
or @weralabaj

src/SqlPersistence/Outbox/OutboxCommandBuilder.cs
```diff
@@ -62,5 +61,24 @@ public static OutboxCommands Build(string tablePrefix, string schema, SqlVariant
...
         }
 
+        static string GetCleanupCommand(SqlVariant sqlVariant, string tableName)
+        {
+            switch (sqlVariant)
+            {
+                case SqlVariant.MsSqlServer:
+                    return $@"
+delete top (@BatchSize) from {tableName}
+where Dispatched = 1
```
weralabaj: > where Dispatched = 1

why not `=true` anymore?

src/SqlPersistence/Outbox/OutboxPersister.cs
```diff
@@ -15,11 +15,13 @@
 class OutboxPersister : IOutboxStorage
 {
     Func<DbConnection> connectionBuilder;
+    int cleanupBatchCount;
     OutboxCommands outboxCommands;
 
-    public OutboxPersister(Func<DbConnection> connectionBuilder, string tablePrefix, string schema, SqlVariant sqlVariant)
+    public OutboxPersister(Func<DbConnection> connectionBuilder, string tablePrefix, string schema, SqlVariant sqlVariant, int cleanupBatchCount = 10000)
```
weralabaj: > cleanupBatchCount 

IMO cleanupBatchSize is a better name, as count might indicate how many batches we want to divide that into

weralabaj: is 10000 just random magic number?

weralabaj:
@SimonCropp few small questions/comments, overall LGTM

src/SqlPersistence/Outbox/OutboxCommandBuilder.cs
SimonCropp: fixed

src/SqlPersistence/Outbox/OutboxPersister.cs
SimonCropp: stolen from the sql transport. yes magic number

SimonCropp: cleanupBatchSize is better. fixed

PR #58, opened 2017-03-17
[SPIKE] Fixed-size outbox

SzymonPobiega:
The spike implement the solution described [here](https://particularsoftware.slack.com/files/szymon.pobiega/F4LED4JGN/Fixed-size_outbox), which is a fixed-size inbox-outbox that does not require periodic cleanup.

The idea is based on decoupling the deduplication from forwarding resulting in two tables: `InboxData` and `OutboxData`. The former is fixed-size and contains message IDs for deduplication while the latter is used for temporarily storing outgoing messages. Because the outbox records are not necessary for deduplication, they can be removed immediately when messages are dispatched.

The spike works only for SQL Server.


SimonCropp:
@SzymonPobiega given how much the codebase has moved forward. should we delete this?

SzymonPobiega:
@SimonCropp I want to come back to it at some point. I should probably move it to my fork. I'll close it and if I need it, I'll restore the branch.

PR #59, opened 2017-03-18
WIP: add cache subscription option

### LINKED ISSUE #56: Cache subscriptions for a short amount of time
Hi,

Please make it possible to cache subscriptions when using Sql Persistence, just like NHibernate: https://github.com/Particular/NServiceBus.NHibernate/pull/33

Thanks,
Jimmy

SimonCropp:
fixes #56

### DIFF
--- a/src/SqlPersistence/Subscription/SubscriptionPersister.cs
+++ b/src/SqlPersistence/Subscription/SubscriptionPersister.cs
@@ -95,7 +95,7 @@ class SubscriptionPersister : ISubscriptionStorage
 
         cacheItem = new CacheItem
         {
-            Stored = DateTimeOffset.UtcNow,
+            Stored = DateTime.UtcNow,
             Subscribers = baseSubscribers.ToList()
         };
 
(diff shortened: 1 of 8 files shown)


src/SqlPersistence/Subscription/SubscriptionPersister.cs
```diff
@@ -66,6 +74,44 @@ public SubscriptionPersister(Func<DbConnection> connectionBuilder, string tableP
...
+        };
+
+        cache.AddOrUpdate(key, s => cacheItem, (s, tuple) => cacheItem);
+
+        return cacheItem.Subscribers;
+    }
+
+    class CacheItem
+    {
+        public DateTimeOffset Stored;
```
SzymonPobiega: Do we need offset part?

SimonCropp: TBH i did not think so. but NH uses it. since we are in memory and always DTC i think change it to datetime?

SzymonPobiega: 👍 for simple `DateTime` and then LGTM

PR #62, opened 2017-03-20
Enable periodic cleanup of expired outbox records

SzymonPobiega:
Wires up the outbox cleaner task using the defaults same as in NHibernate persister:
 * Cleanup interval 1 minute
 * Outbox record life 7 days

### DIFF
--- a/src/SqlPersistence.Tests/Outbox/OutboxPersisterTests.cs
+++ b/src/SqlPersistence.Tests/Outbox/OutboxPersisterTests.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Data.Common;
+using System.Threading;
 using System.Threading.Tasks;
 using NServiceBus.Outbox;
 using NServiceBus.Persistence.Sql.ScriptBuilder;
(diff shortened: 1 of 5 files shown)


SimonCropp:
@SzymonPobiega i did some testing and found a few more issues. so i ported the test from develop to verify. can u QA please?

src/SqlPersistence.Tests/Outbox/OutboxPersisterTests.cs
```diff
@@ -148,4 +149,63 @@ public void StoreAndGet()
         return await persister.Get(messageId, null);
     }
 
+    [Test]
+    public async Task StoreAndCleanup()
+    {
+        using (var connection = await dbConnection.OpenConnection())
+        {
+            for (var i = 0; i < 13; i++)
```
SzymonPobiega: Why 13? Wouldn't 1 be enough?

src/SqlPersistence.Tests/Outbox/OutboxPersisterTests.cs
```diff
@@ -148,4 +149,63 @@ public void StoreAndGet()
...
+            {
+                await Store(i, connection);
+            }
+        }
+
+        await Task.Delay(1000);
```
SzymonPobiega: What do you think about passing `Func<DateTime>` to `OutboxPersister` in `ctror` to make this tests not time-dependent?

src/SqlPersistence.Tests/Outbox/OutboxPersisterTests.cs
SimonCropp: well i wanted more than 1 so multiple would be cleaned up. so any number greater than 1 would do

src/SqlPersistence.Tests/Outbox/OutboxPersisterTests.cs
SimonCropp: happy to. but i think that can be an optimisation on develop?

PR #65, opened 2017-03-24
Outbox support for Oracle

MarcinHoppe:


src/ScriptBuilder/Outbox/Drop_Oracle.sql
```diff
@@ -0,0 +1,28 @@
+﻿declare 
+  tableName varchar2(30) := UPPER(:1) || 'OD';
+  indexName varchar2(30) := tableName || '_IDX';
```
DavidBoike: IX

src/SqlPersistence/Outbox/OutboxCommandBuilder.cs
```diff
@@ -20,11 +20,85 @@ public static OutboxCommands Build(string tablePrefix, string schema, SqlVariant
                 case SqlVariant.MySql:
                     tableName = $"`{tablePrefix}OutboxData`";
                     break;
+                case SqlVariant.Oracle:
+                    tableName = $"\"{tablePrefix}OD\"";
```
DavidBoike: I got burned in the acceptance tests by having the double quotes in here, because then `select from "TableName"` doesn't work because of casing, where `select from TableName` is understood just fine.

MarcinHoppe:
@DavidBoike I applied your comments. Another look would be very much appreciated.

PR #67, opened 2017-03-24
Fix indexes on the Outbox table on SQL Server

### LINKED ISSUE #66: Outbox cleanup does full table scan on SQL Server
### What is the problem?

Currenly the Outbox on SQL Server creates two separate indexes on `Dispatched` and `DispatchedAt` columns.

Execution plan of the cleanup query:

```sql
delete top (@BatchSize) from table
where Dispatched = 'true'
    and DispatchedAt < @DispatchedBefore"
```

Indicates that a table scan is performed. All the other queries only use the primary key index (unique lookup by the primary key).

These two indexes are useless and have negative performance impact (they have to be updated on insert and when marking an Outbox record as dispatched).

### How to solve it?

A single filtered index will resolve it:

```sql
create index table_dispatched_idx on table (DispatchedAt) where Dispatched = 1
```

MarcinHoppe:
Fixes #66

### DIFF
--- a/src/ScriptBuilder/Outbox/Create_MsSqlServer.sql
+++ b/src/ScriptBuilder/Outbox/Create_MsSqlServer.sql
@@ -32,23 +32,6 @@ begin
   declare @createDispatchedAtIndex nvarchar(max);
   set @createDispatchedAtIndex = '
   create index Index_DispatchedAt
-  on ' + @tableName + '(DispatchedAt);';
+  on ' + @tableName + '(DispatchedAt) where Dispatched = 1;';
   exec(@createDispatchedAtIndex);
-end
-
-
-if not exists
-(
-    select *
-    from sys.indexes
-    where
-        name = 'Index_Dispatched' and
-        object_id = object_id(@tableName)
-)
-begin
-  declare @createDispatchedIndex nvarchar(max);
-  set @createDispatchedIndex = '
-  create index Index_Dispatched
-  on ' + @tableName + '(Dispatched);';
-  exec(@createDispatchedIndex);
 end
--- a/src/ScriptBuilder.Tests/OutboxScriptBuilderTest.BuildCreateScript.ForScenario.MsSqlServer.approved.txt
+++ b/src/ScriptBuilder.Tests/OutboxScriptBuilderTest.BuildCreateScript.ForScenario.MsSqlServer.approved.txt
@@ -32,23 +32,6 @@ begin
   declare @createDispatchedAtIndex nvarchar(max);
   set @createDispatchedAtIndex = '
   create index Index_DispatchedAt
-  on ' + @tableName + '(DispatchedAt);';
+  on ' + @tableName + '(DispatchedAt) where Dispatched = 1;';
   exec(@createDispatchedAtIndex);
-end
-
-
-if not exists
-(
-    select *
-    from sys.indexes
-    where
-        name = 'Index_Dispatched' and
-        object_id = object_id(@tableName)
-)
-begin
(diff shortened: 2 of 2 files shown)


MarcinHoppe:
@Particular/sqlpersistence-maintainers The build is green and this is ready for review.

src/ScriptBuilder/Outbox/Create_MsSqlServer.sql
```diff
@@ -32,23 +32,6 @@ begin
...
   exec(@createDispatchedAtIndex);
-end
-
-
-if not exists
-(
-    select *
-    from sys.indexes
-    where
-        name = 'Index_Dispatched' and
```
SimonCropp: if Index_Dispatched is no longer required should it be cleaned up?

MarcinHoppe: It should. Sounds like something that should be addressed in the upgrade guide?

SzymonPobiega:
Apart from Simon's comments, LGTM.

src/ScriptBuilder/Outbox/Create_MsSqlServer.sql
SimonCropp: fair enough. do u want to handle that? if not i can.

MarcinHoppe: Go on. I'll be happy to review.

SimonCropp: ok but it wont happen till monday

MarcinHoppe: Sure, I wouldn't have worked on it any sooner.

SimonCropp: actually this index was added in v2. so no doco needed

SimonCropp:
@MarcinHoppe thanks for the effort of research and PR

SimonCropp:
@DavidBoike sorry given u added those indexes i would appreciate your review of this change

src/ScriptBuilder/Outbox/Create_MsSqlServer.sql
```diff
@@ -32,23 +32,6 @@ begin
   declare @createDispatchedAtIndex nvarchar(max);
   set @createDispatchedAtIndex = '
   create index Index_DispatchedAt
-  on ' + @tableName + '(DispatchedAt);';
+  on ' + @tableName + '(DispatchedAt) where Dispatched = 1;';
```
DavidBoike: So, this is a [filtered index](https://docs.microsoft.com/en-us/sql/relational-databases/indexes/create-filtered-indexes), which was new in SQL 2016, and we [currently support down to SQL 2012](https://docs.particular.net/nservicebus/sql-persistence/?version=sqlpersistence_2#supported-sql-implementations). :frown:

DavidBoike: Ok nevermind, I guess it was [introduced in SQL 2008](https://www.mssqltips.com/sqlservertip/1785/sql-server-filtered-indexes-what-they-are-how-to-use-and-performance-advantages/) and the Microsoft link with its **SQL Server (starting with 2016)** label at the top is full of it.

DavidBoike:
I'm retroactively :+1: - really good use of a filtered index, glad we could take advantage.

SimonCropp:
@DavidBoike thanks for double checking this

PR #68, opened 2017-03-27
remove saga attribute

SimonCropp:
Removes the requirement for the `[SqlSagaAttribute]`

In 2.0 we have already pushed Correlation Property into a abstract property on the `SqlSaga<T>` https://docs.particular.net/nservicebus/upgrades/sqlpersistence-1to2#correlation-property

Given that i think it would be sensible to push both table suffix and Transitional Correlation Property into virtual properties on  `SqlSaga<T>`. and hence remove any requirement for `[SqlSagaAttribute]`

### DIFF
--- a/src/SqlPersistence/Saga/SqlSagaTypeDataReader.cs
+++ b/src/SqlPersistence/Saga/SqlSagaTypeDataReader.cs
@@ -1,37 +1,56 @@
 ﻿using System;
 using System.Reflection;
 using System.Runtime.Serialization;
-using NServiceBus.Persistence.Sql;
 
 static class SqlSagaTypeDataReader
 {
     public static SqlSagaTypeData GetTypeData(Type sagaType)
     {
-        var attribute = sagaType.GetCustomAttribute<SqlSagaAttribute>(false);
-        string tableName = null;
-        string transitionalCorrelationProperty = null;
-        if (attribute != null)
+        var instance = FormatterServices.GetUninitializedObject(sagaType);
+
+        string transitionalCorrelationPropertyName = null;
+        var transitionalCorrelationProperty = GetProperty(sagaType, "TransitionalCorrelationPropertyName");
+        if (transitionalCorrelationProperty != null)
         {
-            tableName = attribute.TableSuffix;
-            transitionalCorrelationProperty = attribute.TransitionalCorrelationProperty;
+            transitionalCorrelationPropertyName = GetPropertyValue(transitionalCorrelationProperty, instance);
         }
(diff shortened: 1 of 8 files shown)


src/SqlPersistence/Saga/SqlSagaTypeDataReader.cs
```diff
@@ -1,37 +1,56 @@
...
+    {
+        property.GetMethod.Invoke(instance, null);
+        return (string) property.GetMethod.Invoke(instance, null);
+    }
+
+    static PropertyInfo GetProperty(Type sagaType, string propertyName)
     {
-        return sagaType.GetProperty("CorrelationPropertyName", BindingFlags.Instance | BindingFlags.GetProperty | BindingFlags.NonPublic | BindingFlags.Public);
+        var propertyInfo = sagaType.GetProperty(propertyName, BindingFlags.Instance | BindingFlags.GetProperty | BindingFlags.NonPublic | BindingFlags.Public);
+        if (propertyInfo != null)
```
SzymonPobiega:
```
if (propertyInfo == null || propertyInfo.DeclaringType != sagaType)
{
   return null;
}
``` 
would be more compact and read better

src/SqlPersistence/Saga/SqlSagaTypeDataReader.cs
```diff
@@ -1,37 +1,56 @@
...
-            TransitionalCorrelationProperty = transitionalCorrelationProperty
+            CorrelationProperty = GetPropertyValue(correlationProperty, instance),
+            TransitionalCorrelationProperty = transitionalCorrelationPropertyName
         };
     }
 
-    static PropertyInfo GetCorrelationProperty(Type sagaType)
+    static string GetPropertyValue(PropertyInfo property, object instance)
+    {
+        property.GetMethod.Invoke(instance, null);
```
SzymonPobiega: Why invoke getter two times?

SzymonPobiega:
Besides two minor comments, I ❤️ it

src/SqlPersistence/Saga/SqlSagaTypeDataReader.cs
SimonCropp: how about `if (propertyInfo?.DeclaringType != sagaType)`?

SzymonPobiega: I think it is not that obvious what this code does if `propertyInfo` is `null` but I can live with it ;-)

weralabaj:
LGTM

PR #75, opened 2017-03-28
Storage adapter adapts to ambient transaction if present

### LINKED ISSUE #74: StorageSession disposes native transaction when enrolled in ambient transaction
## Who's affected

All customers using the persister with a transport that supports DTC such as MSMQ

## Symptoms

When the persister is used with a transport that uses TransactionScope to support DTC operations the following happens

1. The storage adapter returns `null`
1. SynchronizedStorage is requested to open a new session
1. A new connection is created which enlists into the ambient transaction
1. A new native transaction is created
1. When a message fails to be processed the synchronized storage session gets disposed with the indicator that it owns the transaction and disposes the transaction as well as the connection. This will immediately roll back the ambient transaction which makes the message visible in the input queue immediately

## Steps to reproduce

- Use MSMQ with sql persister, set delayed retries and immediate retries to off, count the handler executions with an interlocked, throw an exception in a handler _or_
- Execute `When_custom_policy_always_moves_to_error`

danielmarbach:
Closes #74

### DIFF
--- a/src/SqlPersistence/SynchronizedStorage/StorageAdapter.cs
+++ b/src/SqlPersistence/SynchronizedStorage/StorageAdapter.cs
@@ -1,5 +1,8 @@
+using System;
+using System.Data.Common;
 using System.Data.SqlClient;
 using System.Threading.Tasks;
+using System.Transactions;
 using NServiceBus.Extensibility;
 using NServiceBus.Outbox;
 using NServiceBus.Persistence;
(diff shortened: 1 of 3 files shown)


danielmarbach:
@Particular/sqlpersistence-maintainers please review

src/SqlPersistence/SynchronizedStorage/StorageAdapter.cs
```diff
@@ -23,19 +31,29 @@ public Task<CompletableSynchronizedStorageSession> TryAdapt(OutboxTransaction tr
...
         if (transportTransaction.TryGet(out existingSqlConnection) && transportTransaction.TryGet(out existingSqlTransaction))
         {
             CompletableSynchronizedStorageSession session = new StorageSession(existingSqlConnection, existingSqlTransaction, false, infoCache);
-            return Task.FromResult(session);
+            return session;
         }
+
+        Transaction existingTransaction;
+        // Transport supports DTC and uses TxScope owned by the transport
+        if (transportTransaction.TryGet(out existingTransaction))
```
danielmarbach: I think it should be fairly safe to do it this way since only SQL will add the connection and native tx to the transport transaction which should take precedence over this check here. Thoughts?

danielmarbach:
Targetted against the core 62 branch. Happy to retarget if required

SimonCropp:
@SzymonPobiega i merged this into the 6.2 staging branch. can u give the change a once over before we merge to develop?

src/SqlPersistence/SynchronizedStorage/StorageAdapter.cs
```diff
@@ -23,19 +31,29 @@ public Task<CompletableSynchronizedStorageSession> TryAdapt(OutboxTransaction tr
...
+        {
+            var connection = await connectionBuilder.OpenConnection();
+            CompletableSynchronizedStorageSession session = new StorageSession(connection, NoSqlTransaction, true, infoCache);
```
SzymonPobiega: Should we add explicit call to `Enlist` in order to enlist the new connection in the transaction if auto-enlist is not specified in connection string?

danielmarbach: @SzymonPobiega not a bad idea. I discussed today a few other scenarios with @SimonCropp I will run through those with you in my morning.

Issue #17, opened 2017-02-16
Add Oracle Support


SimonCropp:

DavidBoike:
Completed in 2.1.0

* [Release notes](https://github.com/Particular/NServiceBus.Persistence.Sql/releases/tag/2.1.0)
* [Announcement](https://groups.google.com/forum/#!topic/particularsoftware/XXvdEME7uFo)
* [Connection details in docs](https://docs.particular.net/nservicebus/sql-persistence/#usage-oracle)
* [Oracle-specific caveats](https://docs.particular.net/nservicebus/sql-persistence/oracle-caveats) dealing mostly with table length restrictions

PR #76, opened 2017-03-28
Oracle support

DavidBoike:
Connects to #17 

First commit is the persistence implementation. Second commit is the acceptance test package including all the sources, which are basically non-reviewable, and third commit configures the acceptance tests.

Currently acceptance tests pass locally.

~~The Oracle saga persister unit tests are broken after the rebase onto develop because all the SagaMetadata being loaded results in the metadata for `SagaWithWeirdCharactersಠ_ಠ` being loaded for _every_ test, and Oracle has to throw because it does not support non-ASCII characters. I'm not sure what to do about that - open to suggestions.~~ Now resolved.

@Particular/sqlpersistence-maintainers @MarcinHoppe


src/ScriptBuilder/Subscription/Create_Oracle.sql
```diff
@@ -0,0 +1,29 @@
...
+        , ENDPOINT VARCHAR2(200) NOT NULL 
+        , PERSISTENCEVERSION VARCHAR2(23) 
+        , CONSTRAINT ' || tableName || '_PK PRIMARY KEY 
+          (
+            MESSAGETYPE 
+          , SUBSCRIBER 
+          )
+          ENABLE 
+        ) 
+        ORGANIZATION INDEX';
```
MarcinHoppe: Does this help vs regular table with an index? Do we know what the impact on the execution plan is?

src/SqlPersistence.Tests/Saga/OracleSagaPersisterTests.cs
```diff
@@ -0,0 +1,30 @@
...
+    public OracleSagaPersisterTests() : base(BuildSqlVariant.Oracle, null)
+    {
+    }
+
+    protected override Func<DbConnection> GetConnection()
+    {
+        return () =>
+        {
+            var connection = OracleConnectionBuilder.Build();
+            connection.Open();
```
MarcinHoppe: Is this explicit call to `Open` needed here? Other test fixtures seem to be able to get away without one.

src/ScriptBuilder/Subscription/Create_Oracle.sql
DavidBoike: I admit I'm making assumptions based on my knowledge of the data model as it pertains to SQL Server and my understanding of SQL clustered indexes vs. Oracle index-organized tables. I don't know (yet) how to look at the execution plan but it's something we should definitely look at.

src/SqlPersistence.Tests/Saga/OracleSagaPersisterTests.cs
DavidBoike: It matches the MsSQL/MySQL saga test classes. Just based upon how the SagaPersisterTests are set up in comparison to the other persisters.

src/ScriptBuilder/Subscription/Create_Oracle.sql
MarcinHoppe: I added a PoA item to examine execution plans. Let's cover this index organized table there.

src/SqlPersistence.Tests/Saga/OracleSagaPersisterTests.cs
MarcinHoppe: Fair enough.

DavidBoike:
The acceptance tests have saga names that are > 27 characters. Found a way to slice those off without requiring changes to the acceptance tests themselves. I figure in the acceptance tests our utmost care is the behavior of the tests, where in real life you'd have to go fix your saga names to fit within your chosen persistence.

I did have to change 2 tests related to timeouts, which aren't even invoking the timeout persister (directly dispatching a message with 2s timeout because it's set to expire "soon" but then still demanding that the delay be longer than 2s) so I believe they are invalid tests in core.

And hey look, the tests are all green now. :smile:

DavidBoike:
@Particular/sqlpersistence-maintainers I removed WIP on this PR. Please have a look. Going to be starting on a doco PR today or tomorrow.

SzymonPobiega:
@MarcinHoppe would you be able to squash all the implementation commits so that there is singe one for the code and a single one for the acceptance tests?

MarcinHoppe:
I squashed but it looked worse and I believe it would be harder to review. @SzymonPobiega and I agreed to leave this PR as is.

src/ScriptBuilder/Saga/Writers/OracleSagaScriptWriter.cs
```diff
@@ -0,0 +1,209 @@
...
+
+    public OracleSagaScriptWriter(TextWriter textWriter, SagaDefinition saga)
+    {
+        writer = textWriter;
+        this.saga = saga;
+        if (saga.TableSuffix.Length > 27)
+        {
+            throw new Exception($"Saga '{saga.TableSuffix}' contains more than 27 characters, which is not supported by SQL persistence using Oracle. Either disable Oracle script generation using the SqlPersistenceSettings assembly attribute, shorten the name of the saga, or specify an alternate table name using the SqlSagaAttribute's 'tablePrefix' parameter.");
+        }
+        if (Encoding.UTF8.GetBytes(saga.TableSuffix).Length != saga.TableSuffix.Length)
```
SzymonPobiega: I think checking each `char` of the string if it is higher than 255 (or 127, not sure which set Oracle supports) would be cleaner and would be equivalent.

src/ScriptBuilder/Saga/SagaScriptBuilder.cs
```diff
@@ -65,6 +65,9 @@ public static void BuildCreateScript(SagaDefinition saga, BuildSqlVariant sqlVar
 
             WriteComment(writer, "PurgeObsoleteProperties");
             sqlVariantWriter.WritePurgeObsoleteProperties();
+
+            WriteComment(writer, "CreateComplete");
```
weralabaj: I think something like "Complete saga script" will be clearer when reading, given that the other two don't have it.

BTW we could review what common pieces we really have at some point, it gets a bit complicated and messy (e.g. Initialize() is empty for MSSQL, WriteTableNameVariable() empty for Oracle, etc.).

src/ScriptBuilder/Saga/Writers/OracleCorrelationPropertyTypeConverter.cs
```diff
@@ -0,0 +1,26 @@
...
+    public static class OracleCorrelationPropertyTypeConverter
+    {
+        public static string GetColumnType(CorrelationPropertyType propertyType)
+        {
+            switch (propertyType)
+            {
+                case CorrelationPropertyType.DateTime:
+                    return "TIMESTAMP";
+                case CorrelationPropertyType.String:
+                    return "NVARCHAR2(200)";
```
weralabaj: Are we documenting this limit (200)? Why this number?

src/SqlPersistence/Saga/RuntimeSagaInfo.cs
```diff
@@ -62,9 +66,23 @@ class RuntimeSagaInfo
...
+                FillParameter = ParameterFiller.Fill;
                 break;
             case SqlVariant.MySql:
                 TableName = $"`{tablePrefix}{tableSuffix}`";
+                FillParameter = ParameterFiller.Fill;
+                break;
+            case SqlVariant.Oracle:
+                if (tableSuffix.Length > 27)
+                {
+                    throw new Exception($"Saga '{tableSuffix}' contains more than 27 characters, which is not supported by SQL persistence using Oracle. Either disable Oracle script generation using the SqlPersistenceSettings assembly attribute, shorten the name of the saga, or specify an alternate table name using the SqlSagaAttribute's 'tablePrefix' parameter.");
```
SzymonPobiega: Is there a way to centralize these checks somehow in a definition of a SQL dialect? I am pretty sure I've seen this two times already ;-)

src/ScriptBuilder/Saga/Writers/OracleCorrelationPropertyTypeConverter.cs
```diff
@@ -0,0 +1,26 @@
...
+                case CorrelationPropertyType.Int:
+                    return "NUMBER(19)";
+                case CorrelationPropertyType.Guid:
+                    return "VARCHAR2(38)";
```
weralabaj: Not 36?

src/ScriptBuilder/Saga/Writers/OracleSagaScriptWriter.cs
```diff
@@ -0,0 +1,214 @@
...
+    SagaDefinition saga;
+    string tableName;
+
+    public OracleSagaScriptWriter(TextWriter textWriter, SagaDefinition saga)
+    {
+        writer = textWriter;
+        this.saga = saga;
+        if (saga.TableSuffix.Length > 27)
+        {
+            throw new Exception($"Saga '{saga.TableSuffix}' contains more than 27 characters, which is not supported by SQL persistence using Oracle. Either disable Oracle script generation using the SqlPersistenceSettings assembly attribute, shorten the name of the saga, or specify an alternate table name by overriding the SqlSaga's TableSuffix property.");
```
weralabaj: > not supported by SQL persistence using Oracle

Or by Oracle itself?

> Either disable Oracle script generation using the SqlPersistenceSettings assembly attribute

I'm not clear on how does that solve the problem?

src/ScriptBuilder/Saga/Writers/OracleSagaScriptWriter.cs
```diff
@@ -0,0 +1,214 @@
...
+        }
+        if (Encoding.UTF8.GetBytes(saga.TableSuffix).Length != saga.TableSuffix.Length)
+        {
+            throw new Exception($"Saga '{saga.TableSuffix}' contains non-ASCII characters, which is not supported by SQL persistence using Oracle. Either disable Oracle script generation using the SqlPersistenceSettings assembly attribute, change the name of the saga, or specify an alternate table name by overriding the SqlSaga's TableSuffix property.");
+        }
```
weralabaj: as above

SzymonPobiega:
Apart from my comments above, LGTM.

src/SqlPersistence/Timeout/TimeoutCommandBuilder.cs
```diff
@@ -18,6 +18,8 @@ public static TimeoutCommands Build(SqlVariant sqlVariant, string tablePrefix, s
                     return BuildMySqlCommands($"`{tablePrefix}TimeoutData`");
                 case SqlVariant.MsSqlServer:
                     return BuildSqlServerCommands($"[{schema}].[{tablePrefix}TimeoutData]");
+                case SqlVariant.Oracle:
+                    return BuildOracleCommands($"{tablePrefix}TO");
```
weralabaj: Why TO not TD?

src/ScriptBuilder/Saga/SagaScriptBuilder.cs
weralabaj: I think something like "Complete saga script" will be clearer when reading, given that the other two don't have it.

BTW we could review what common pieces we really have at some point, it gets a bit complicated and messy (e.g. Initialize() is empty for MSSQL, WriteTableNameVariable() empty for Oracle, etc.).

src/ScriptBuilder/Saga/Writers/OracleCorrelationPropertyTypeConverter.cs
weralabaj: Are we documenting this limit (200)? Why this number?

src/ScriptBuilder/Saga/Writers/OracleCorrelationPropertyTypeConverter.cs
weralabaj: Not 36?

src/ScriptBuilder/Saga/Writers/OracleSagaScriptWriter.cs
```diff
@@ -0,0 +1,214 @@
...
+    string tableName;
+
+    public OracleSagaScriptWriter(TextWriter textWriter, SagaDefinition saga)
+    {
+        writer = textWriter;
+        this.saga = saga;
+        if (saga.TableSuffix.Length > 27)
+        {
+            throw new Exception($"Saga '{saga.TableSuffix}' contains more than 27 characters, which is not supported by SQL persistence using Oracle. Either disable Oracle script generation using the SqlPersistenceSettings assembly attribute, shorten the name of the saga, or specify an alternate table name by overriding the SqlSaga's TableSuffix property.");
+        }
```
weralabaj: > not supported by SQL persistence using Oracle

Or by Oracle itself? I think it's a bit confusing from the user perspective that they need to be aware that the overall Oracle limit is 30 and in saga we take 3 chars, but in other tables we might take 2. Could we say in all places that the overall limit is 30 and show the full name? People will figure out probably which part is "fixed", but won't have to remember a few different numbers.

> Either disable Oracle script generation using the SqlPersistenceSettings assembly attribute

I'm not clear on how does that solve the problem?

src/ScriptBuilder/Saga/Writers/OracleSagaScriptWriter.cs
weralabaj: as above

src/SqlPersistence/Timeout/TimeoutCommandBuilder.cs
weralabaj: Why TO not TD?

src/SqlPersistence/Installer.cs
```diff
@@ -111,10 +141,19 @@ Task InstallSagas(string scriptDirectory, DbConnection connection, DbTransaction
...
-        return connection.ExecuteTableCommand(
-            transaction: transaction,
-            scripts: sagaScripts,
-            tablePrefix: tablePrefix,
-            schema: schema);
+        if (sqlVariant == SqlVariant.Oracle)
+        {
+            return connection.ExecuteTableCommand(
+                transaction: transaction,
+                scripts: sagaScripts);
```
weralabaj: Didn't we have an optional table prefix for sagas too?

MarcinHoppe: It is not used on Oracle and documented here: https://github.com/Particular/docs.particular.net/blob/sql-oracle/nservicebus/sql-persistence/oracle-caveats.md#sagas.

src/ScriptBuilder/Subscription/Create_Oracle.sql
DavidBoike: We looked at execution plans and they looked good. I consider this resolved.

src/ScriptBuilder/Saga/Writers/OracleSagaScriptWriter.cs
DavidBoike: This can't be centralized because it's needed both in SqlPersistence and ScriptBuilder and they have no common dependency. Given that and it's not a hot path, isn't it better to have a simple one-liner rather than call out to another method? I also think it conveys intent better than `saga.TableSuffix.Any(ch => (int)ch > 127)`.

src/SqlPersistence/Saga/RuntimeSagaInfo.cs
DavidBoike: This can't be centralized because it's needed both in SqlPersistence and ScriptBuilder and they have no common dependency. I agree conceptually though.

src/ScriptBuilder/Saga/Writers/OracleCorrelationPropertyTypeConverter.cs
DavidBoike: Same as MySQL, I believe it's intended to allow for { } braces before and after the guid. Will be documented by the Oracle Scripts page.

src/ScriptBuilder/Saga/Writers/OracleCorrelationPropertyTypeConverter.cs
DavidBoike: Same as MySQL and MSSQL, and I believe inherited from NHibernate. Some rational limit on how long an OrderId (or whatever the correlation property would be) has to be set. Will be documented by the Oracle Scripts page.

src/ScriptBuilder/Saga/Writers/OracleSagaScriptWriter.cs
DavidBoike: Oracle requires <= 30. We are going to take 3 for the addition of `_PK` for the primary key name. I was trying to tell the user exactly what their limitation is rather than explain all the whys, which should be covered by doco. Open to better suggestions though.

> Either disable Oracle script generation using the SqlPersistenceSettings assembly attribute

It could be that you are using MSSQL/MySQL and so far you have short saga names, and you haven't disabled Oracle script generation. Now you add a 28 character saga name - you don't really need to shorten your saga name for MSSQL, you just need to stop generating the Oracle scripts you don't need.

src/SqlPersistence/Timeout/TimeoutCommandBuilder.cs
DavidBoike: Time Outs? Was trying to use specific acronyms without resorting to D for Data in all cases, but...

* Time Outs = TO
* Subscription Storage = SS
* Out Box = OB .... hmmmm, OB usually means something else to people

DavidBoike:
@weralabaj @SzymonPobiega I responded to a bunch of comments. If you take a look at those I'll see if I can rebase one more time on top of develop to make it easier for a final pass.

SzymonPobiega:
LGTM

src/TestHelper/SqlHelpers.cs
```diff
@@ -12,13 +12,9 @@ public static void ExecuteCommand(this DbConnection connection, string script, s
             {
                 command.CommandText = script;
                 command.AddParameter("tablePrefix", $"{tablePrefix}_");
-                if (schema == null)
+                if (command is System.Data.SqlClient.SqlCommand)
```
SimonCropp: isnt this a bit leaky?

src/SqlPersistence/CommandWrapper.cs
```diff
@@ -0,0 +1,70 @@
...
+
+    public DbTransaction Transaction
+    {
+        get { return command.Transaction; }
+        set { command.Transaction = value; }
+    }
+
+    public virtual void AddParameter(string name, object value)
+    {
+        var parameter = command.CreateParameter();
```
SimonCropp: doesnt making this an extension method on `DbCommand` render this class redundant?

DavidBoike: I would love that, but not sure how it could work. This is just the default implementation - OracleCommandWrapper inherits from this and which one you get is determined by the CommandBuilder based on your sqlVariant. The OracleCommandWrapper fills parameters completely differently.

I'm not sure what we could switch on in an extension method, since we don't have a reference to the Oracle components, we can't very well check what kind of command this is, without resorting to GetType() and string comparisons.

OracleCommandWrapper also sets BindByName on the OracleCommand (by reflection, it's gross) on construction.

src/TestHelper/SqlHelpers.cs
DavidBoike: In what way? Currently isn't SQL Server the only engine that supports/needs a schema? MySQL doesn't but it can tolerate extra unused parameters. Oracle throws if you give it parameters you're not using.

Would you prefer passing in the sqlVariant to these? Or making SqlHelpers a non-static class that takes a sqlVariant in its constructor?

src/ScriptBuilder/Saga/Writers/OracleSagaScriptWriter.cs
```diff
@@ -0,0 +1,212 @@
...
+        if (name.Length > 30)
+        {
+            name = name.Substring(0, 30);
+        }
+        return name;
+    }
+
+    // Generates a 30ch index name like "SAGAIDX_66462BF46C9DCB70FD257D" based on
+    // SHA1 hash of saga name and correlation property name
+    string CreateSagaIndexName(string sagaName, string correlationPropertyName)
```
SimonCropp: i dont think we should default this. i would prefer this was a required config with us suggesting an approach in doco. the reason is every case where i have seen us own a shortening algorithm it has bitten us in the long run

SimonCropp: also isnt this theoretically a problem in othr DBs as well?

src/ScriptBuilder/Subscription/Create_Oracle.sql
```diff
@@ -0,0 +1,29 @@
+﻿declare
+  tableName varchar2(30) := UPPER(:1) || 'SS';
```
SimonCropp: why UPPER?

SimonCropp:
SQL Server and MySql encode the table names. eg

    declare @tableName nvarchar(max) = '[' + @schema + '].[' + @tablePrefix + 'TimeoutData]';

so thinks like spaces and dots wont break things

has that been considered for oracle?

SimonCropp:
there are a bunch of places screaming caps are used. eg `SAGAIDX_`. can we move over to lower or camel casing to be consistent with mysql and sql server

SimonCropp:
can u add an oracle integration project here https://github.com/Particular/NServiceBus.Persistence.Sql/tree/oracle/src/Integration

MarcinHoppe:
@SimonCropp Oracle project added to Integration.

src/ScriptBuilder/Saga/Writers/OracleSagaScriptWriter.cs
MarcinHoppe: Perhaps I lack historical insight and context here but what are the problems? I know we use hashing in a couple of places (including NH persister ATTs) but I am not aware of problems this causes.

Limit is 64 for MySQL and 128 for SQL Server. Theoretically it could be a problem on these RDBMSs as well.

We could consider creating an API to control this along with guidance in docs but I'd much rather do this as a separate issue.

Thoughts?

SimonCropp: > We could consider creating an API to control this along with guidance in docs but I'd much rather do this as a separate issue

Happy with that

MarcinHoppe:
@SimonCropp Screaming caps seem to be a thing that people using Oracle are well used to. By default Oracle treats all object names as uppercase, unless you explicitly quote them using `"`. When you quote, table (and column, etc.) names suddenly become case sensitive and you have to use `"` in **every** query and never ever miss capitalization of a single letter.

That is the reason we chose to follow the accepted convention in the Oracle community (anecdotal, but still) and do not quote identifiers. We reasoned that our primary target audience for the generated scripts are people who are used to those conversions.

@Particular/sqlpersistence-maintainers Please indicate which items need more work. The code has been rebased, the build is green and we think it is good to be merged to `develop`.

SimonCropp:
the problem with "do not quote identifiers" is if they do use some character that is invalid they will get a fairly hard to debug script exception

MarcinHoppe:
This should be fairly easy to spot in SQL*Plus or SQL Developer. My understanding was that this is the use case we are optimizing for.

DavidBoike:
The thing that gets me is with SQL Server, we quote everything with [ and ] but if the user sticks to a no-quotes-necessary name, they are free to interact with it and query it for management/maintenance purposes with the simpler syntax. If they throw a space in there, then that's on them and they own it and have to quote it with [ and ] if they ever want to talk to it.

In Oracle, if we "QuotedForSafety" but had mixed case, even if they aren't picking funny characters we've obligated them into always using quotes as well, which is probably annoying if they're a WE_ROLL_WITH_UPPERS type of a shop.

So, what if we "QUOTED_FOR_SAFETY" but "ROLL_WITH_UPPERS"? That would give the same experience as with SQL Server, as long as we're OK with a bit of screaming.

SimonCropp:
> So, what if we "QUOTED_FOR_SAFETY" but "ROLL_WITH_UPPERS"? That would give the same experience as with SQL Server, as long as we're OK with a bit of screaming.

i think that is reasonable

SzymonPobiega:
👍 to what @DavidBoike said.

MarcinHoppe:
@DavidBoike @SzymonPobiega @SimonCropp Just so I understand: the idea is to keep converting all object names to UPPER (as we do now in scripts) and put them in `""` for safety?

SzymonPobiega:
@MarcinHoppe the build server has been blocked by SC so it takes ages to get to verify this PR. I haven't forgotten about it and will merge when green.

SzymonPobiega:
🎉 Thanks @DavidBoike & @MarcinHoppe ❤️

PR #78, opened 2017-03-29
Throws when new transaction scope is created in the pipeline

SzymonPobiega:
As discussed with Daniel, the adapter throws if both transport and custom TS are present. Otherwise it uses the one scope that is present. It also explicitly enlists the created connection in the ambient scope in case enlisting is not enable in connection string.

One thing that I am not sure is if all the database drivers we want to target support enlisting and do they do it. Surely not all support DTC-managed transactions.

### DIFF
--- a/src/MsSqlAcceptanceTests/When_creating_new_transaction_scope_in_the_pipeline.cs
+++ b/src/MsSqlAcceptanceTests/When_creating_new_transaction_scope_in_the_pipeline.cs
@@ -1,0 +1,111 @@
+﻿namespace NServiceBus.AcceptanceTests.Sagas
+{
+    using System;
+    using System.Linq;
+    using System.Threading.Tasks;
+    using System.Transactions;
+    using AcceptanceTesting;
+    using EndpointTemplates;
+    using NServiceBus.Pipeline;
+    using NUnit.Framework;
+    using Persistence.Sql;
+
+    public class When_creating_new_transaction_scope_in_the_pipeline : NServiceBusAcceptanceTest
+    {
+        [Test]
+        public async Task Should_fail_when_creating_synchronized_storage_session()
+        {
+            var context = await Scenario.Define<Context>()
+                .WithEndpoint<Endpoint>(b => b
+                    .When(async session =>
+                    {
+                        var id = Guid.NewGuid();
+                        await session.SendLocal(new StartSagaMessage
+                        {
+                            SomeId = id
+                        });
+                    }).DoNotFailOnErrorMessages())
+                .Done(c => c.FailedMessages.Any())
+                .Run();
+
+            var exceptionMessage = context.FailedMessages.First().Value.First().Exception.Message;
+            StringAssert.StartsWith("A TransctionScope has been opened in the current context overriding the one created by the transport.", exceptionMessage);
+        }
+
+        public class Context : ScenarioContext
+        {
+            public bool SecondMessageReceived { get; set; }
+            public Guid FirstSagaId { get; set; }
+            public Guid SecondSagaId { get; set; }
+        }
+
+        public class Endpoint : EndpointConfigurationBuilder
+        {
+            public Endpoint()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    c.Pipeline.Register(new TransactionScopeBehavior(), "Creates a new transaction scope");
+                    c.Pipeline.Register(new SimulateFailureBehavior(), "Simulates failure before committing transport transaction");
+                });
+            }
+
+            class TransactionScopeBehavior : Behavior<IIncomingLogicalMessageContext>
+            {
+                public override async Task Invoke(IIncomingLogicalMessageContext context, Func<Task> next)
+                {
+                    using (var scope = new TransactionScope(TransactionScopeOption.RequiresNew, TransactionScopeAsyncFlowOption.Enabled))
+                    {
+                        await next().ConfigureAwait(false);
+                        scope.Complete();
+                    }
+                }
+            }
+
+            class SimulateFailureBehavior : Behavior<IIncomingPhysicalMessageContext>
+            {
+                public override async Task Invoke(IIncomingPhysicalMessageContext context, Func<Task> next)
+                {
+                    await next().ConfigureAwait(false);
+                    throw new SimulatedException();
+                }
+            }
+
+            public class TestSaga12 : SqlSaga<TestSagaData12>, IAmStartedByMessages<StartSagaMessage>
+            {
+                protected override string CorrelationPropertyName => nameof(TestSagaData12.SomeId);
+
+                public Context TestContext { get; set; }
+
+                public Task Handle(StartSagaMessage message, IMessageHandlerContext context)
+                {
+                    Data.Counter++;
+                    return Task.FromResult(0);
+                }
+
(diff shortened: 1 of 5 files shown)


SzymonPobiega:
Once accepted, I'll copy the code to NHibernate because it needs similar check.

src/SqlPersistence/SynchronizedStorage/StorageAdapter.cs
```diff
@@ -43,11 +43,21 @@ public Task<CompletableSynchronizedStorageSession> TryAdapt(OutboxTransaction tr
...
+        Transaction transportTx;
         // Transport supports DTC and uses TxScope owned by the transport
-        if (transportTransaction.TryGet(out existingTransaction))
+        transportTransaction.TryGet(out transportTx);
+        if (transportTx != null && Transaction.Current != null && !transportTx.Equals(Transaction.Current))
+        {
+            throw new Exception("A TransctionScope has been opened in the current context overriding the one created by the transport. " 
+                + "This setup can result in insonsistent data because operations done via connections enlisted in the context scope won't be committed "
+                + "atomically with the receive transaction. If you wish to manually control the TransactionScope in the pipeline switch the transport transaction mode "
+                + "to values lower than 'TransactionScope'.");
```
danielmarbach: Should we use nameof(TheEnumWhichHasTheMode) here?

SzymonPobiega: Fixed :)

src/MsSqlAcceptanceTests/When_creating_new_transaction_scope_in_the_pipeline.cs
```diff
@@ -0,0 +1,111 @@
...
+
+                protected override void ConfigureMapping(IMessagePropertyMapper mapper)
+                {
+                    mapper.ConfigureMapping<StartSagaMessage>(m => m.SomeId);
+                }
+            }
+
+            public class TestSagaData12 : IContainSagaData
+            {
+                public virtual Guid SomeId { get; set; }
```
SimonCropp: why virtual?

SzymonPobiega: Copy-paste from some generic saga test. I'll remove.

SzymonPobiega: I've cleaned up some other copy-paste junk.

src/SqlPersistence/SynchronizedStorage/StorageAdapter.cs
```diff
@@ -43,11 +44,21 @@ public Task<CompletableSynchronizedStorageSession> TryAdapt(OutboxTransaction tr
...
         }
 
-        Transaction existingTransaction;
+        Transaction transportTx;
         // Transport supports DTC and uses TxScope owned by the transport
-        if (transportTransaction.TryGet(out existingTransaction))
+        transportTransaction.TryGet(out transportTx);
+        if (transportTx != null && Transaction.Current != null && !transportTx.Equals(Transaction.Current))
+        {
+            throw new Exception("A TransctionScope has been opened in the current context overriding the one created by the transport. "
```
SimonCropp: TransctionScope > TransactionScope

SimonCropp: insonsistent > inconsistent

SimonCropp: "If you wish to manually control the TransactionScope" > "To manually control the TransactionScope"

src/SqlPersistence/SynchronizedStorage/StorageAdapter.cs
```diff
@@ -43,11 +44,21 @@ public Task<CompletableSynchronizedStorageSession> TryAdapt(OutboxTransaction tr
             return session;
         }
 
-        Transaction existingTransaction;
+        Transaction transportTx;
         // Transport supports DTC and uses TxScope owned by the transport
-        if (transportTransaction.TryGet(out existingTransaction))
+        transportTransaction.TryGet(out transportTx);
+        if (transportTx != null && Transaction.Current != null && !transportTx.Equals(Transaction.Current))
```
SimonCropp: cant do `if (transportTransaction.TryGet(out transportTx) &&....`  ?

SimonCropp: why not `transportTx !=Transaction.Current`  ?

src/SqlPersistence/SynchronizedStorage/StorageAdapter.cs
```diff
@@ -43,11 +44,21 @@ public Task<CompletableSynchronizedStorageSession> TryAdapt(OutboxTransaction tr
...
+        {
+            throw new Exception("A TransctionScope has been opened in the current context overriding the one created by the transport. " 
+                + "This setup can result in insonsistent data because operations done via connections enlisted in the context scope won't be committed "
+                + "atomically with the receive transaction. If you wish to manually control the TransactionScope in the pipeline switch the transport transaction mode "
+                + $"to values lower than '{nameof(TransportTransactionMode.TransactionScope)}'.");
```
SimonCropp: shouldnt it be `{nameof(TransportTransactionMode)}.{nameof(TransportTransactionMode.TransactionScope)}` ?

src/SqlPersistence/SynchronizedStorage/StorageAdapter.cs
```diff
@@ -43,11 +44,21 @@ public Task<CompletableSynchronizedStorageSession> TryAdapt(OutboxTransaction tr
...
+        }
+        var ambientTransaction = transportTx ?? Transaction.Current;
```
SimonCropp: looking at the decompilation of `Transaction.Current` it might be better to assign it to a local variable

src/SqlPersistence/SynchronizedStorage/StorageAdapter.cs
SzymonPobiega: I don't see `!=` defined in `Transaction` so I thought it would be safer to call `Equals` because equals is overridden and refers to *transaction hash*.

src/SqlPersistence/SynchronizedStorage/StorageAdapter.cs
SzymonPobiega: I think it would be too verbose. The name of the field should be enough to fine the enum.

SimonCropp: ok

SzymonPobiega:
@danielmarbach @SimonCropp I've fixed the ATs by switching the MySQL tests to run MSMQ in native transaction mode.

What I am a bit worried about is that the exception which user got was not very descriptive: `System.NotSupportedException: MySQL Connector/Net does not currently support distributed transactions.` in the point it has been thrown. Should we wrap it in our custom exception?

danielmarbach:
@SzymonPobiega the drawback of wrapping could be that it is harder to detect the actual root cause exception. Because if you wrap the exception you have to look at the inner exception to see what actually happened. By raising directly the NotSupportedException one could immediately google that. I'd vote for not wrapping but I'm on the fence

PR #79, opened 2017-03-29
[SPIKE] Support Saga<T> with unmodified Core Acceptance Tests

DavidBoike:
Related to https://github.com/Particular/PlatformDevelopment/issues/1188


SimonCropp:
we actually want the acceptance tests to run against sqlsaga since that is the code that runs in production.

PR #82, opened 2017-04-13
Reading fields instead of properties for settings attribute

danielmarbach:
The settings attribute reading is broken. The settings attribute reading defines public fields and not properties. Therefore accessing the properties will not return anything. This seems to be broken in latest release as well (in fact it never worked)


SimonCropp:
R u sure? Note we use fielder https://github.com/Particular/NServiceBus.Persistence.Sql/blob/develop/src/SqlPersistence/FodyWeavers.xml#L3

danielmarbach:
Aah! Magic! Sorry missed that

PR #83, opened 2017-04-14
check conn string for enlist props

SimonCropp:
the [enlist change](https://github.com/Particular/NServiceBus.Persistence.Sql/commit/1c09d6dd6d99bbe3d589521724cc30297ec80569) broke mysql. discovered by smoke testing.

results in a `MySQL Connector/Net does not currently support distributed transactions.`

i dont think we should enlist if the connection string has it toggled off

@SzymonPobiega @danielmarbach thoughts?

also any ideas how to add a test this?

### DIFF
--- a/src/SqlPersistence/SynchronizedStorage/StorageAdapter.cs
+++ b/src/SqlPersistence/SynchronizedStorage/StorageAdapter.cs
@@ -15,11 +15,42 @@ class StorageAdapter : ISynchronizedStorageAdapter
 
     SagaInfoCache infoCache;
     Func<DbConnection> connectionBuilder;
+    bool enlistInAmbientTransaction = true;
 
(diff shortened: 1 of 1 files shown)


SimonCropp:
discussed with @SzymonPobiega. the answer is for mysql users to run in a lower transaction mode which he will document.

MarcinHoppe:
@SimonCropp Should this go into 2.1 milestone? It was not released in 2.0, was it?

SimonCropp:
@MarcinHoppe this was not merged. so no milestone

MarcinHoppe:
@SimonCropp Dammit. I am blind even with glasses. Classic PEBCAK case :).

SimonCropp:
@MarcinHoppe  it happens to us all

PR #99, opened 2017-05-22
[wip] Acceptance tests to test special chars in headers

timbussmann:
raising this as a PR to let the tests run on the different SQL servers. Feel free to pull this in or to close this once the tests ran (will be part of the acceptance tests package with the next version anyway)


weralabaj:
MySQL tests pass locally on my machine.

weralabaj:
I think I found the culprit, locally all those are utf8, most likely that's why locally tests pass. Question is - should we do anything apart from documenting this fact? Or just raise an issue for now for persistence maintainers (which include me) and let them (us) handle it? ;)

variables

weralabaj:
I'll open an issue to follow up on this (probably just document the behavior).

PR #102, opened 2017-05-24
Remove table name appendage.

### LINKED ISSUE #101: Funny SQL for MSSQL/MySQL GetSubscribers command
Incorrect string interpolation is resulting in the following GetSubscribers command for Microsoft SQL Server and MySQL, in which the `SubscriptionData` suffix is repeated after the quoted table name.

From the [MS SQL Server Scripts](https://docs.particular.net/nservicebus/sql-persistence/sqlserver-scripts#run-time-subscription-getsubscribers):

```
select distinct Subscriber, Endpoint
from [dbo].[EndpointNameSubscriptionData]SubscriptionData
where MessageType in (@type0)
```

From [MySQL Scripts](https://docs.particular.net/nservicebus/sql-persistence/mysql-scripts#run-time-subscription-getsubscribers):

```
select distinct Subscriber, Endpoint
from `EndpointNameSubscriptionData`SubscriptionData
where MessageType in (@type0)
```

The reason it works (and all unit/acceptance tests pass) is that the extra token is understood by both engines as a table alias, so the query is technically valid and returns the correct results.

Correcting this is a minor issue that can probably be rolled into the next patch or minor.

Note that as Oracle is a separate case (due to the parameter syntax and need for shorter table name) the [Oracle command is not affected](https://docs.particular.net/nservicebus/sql-persistence/oracle-scripts#run-time-subscription-getsubscribers):

```
select distinct Subscriber, Endpoint
from "ENDPOINTNAMESS"
where MessageType in (:type0)
```

DavidBoike:
Fixes #101

### DIFF
--- a/src/SqlPersistence/Subscription/SubscriptionCommandBuilder.cs
+++ b/src/SqlPersistence/Subscription/SubscriptionCommandBuilder.cs
@@ -176,7 +176,7 @@ where MessageType in (";
 
                     var getSubscribersPrefix = $@"
 select distinct Subscriber, Endpoint
-from {tableName}SubscriptionData
+from {tableName}
 where MessageType in (";
 
                     return messageTypes =>


DavidBoike:
@Particular/sqlpersistence-maintainers as mentioned in #101 I'm not sure this merits a release. Can we just merge this to develop and roll it into the next patch/minor?

SzymonPobiega:
@DavidBoike I agree, we don't need to release.

PR #103, opened 2017-05-29
Fixed TimeoutPersister to align with other persistences

SzymonPobiega:
* Fixed the poll query to be exclusive in the lower end: `(startSlice, now]`
 * Added a clean-up mode to the persister (by default triggered every 2 minutes) which returns timeouts from the beginning of time

The clean-up internal is configurable but there is no API for it. We don't expect the need to play with it.


SzymonPobiega:

PR #107, opened 2017-06-12
Retrying subscribe/unsubscribe operations immediately in case of exceptions

SzymonPobiega:
Also using `TABLOCK` hint in the subscribe query

### DIFF
--- a/src/SqlPersistence/Subscription/SubscriptionPersister.cs
+++ b/src/SqlPersistence/Subscription/SubscriptionPersister.cs
@@ -5,19 +5,15 @@ using System.Data.Common;
 using System.Linq;
 using System.Threading.Tasks;
 using NServiceBus.Extensibility;
+using NServiceBus.Logging;
 using NServiceBus.Persistence.Sql;
 using NServiceBus.Unicast.Subscriptions;
 using NServiceBus.Unicast.Subscriptions.MessageDrivenSubscriptions;
+
 #pragma warning disable 618
 
 class SubscriptionPersister : ISubscriptionStorage
 {
-    Func<DbConnection> connectionBuilder;
-    TimeSpan? cacheFor;
-    SubscriptionCommands subscriptionCommands;
-    public ConcurrentDictionary<string, CacheItem> Cache;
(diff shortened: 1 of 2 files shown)


SzymonPobiega:
@SimonCropp @weralabaj how about we change it to a normal `if exists` query using `serializable` isolation level. I guess it would be easier to understand than the overcomplicated syntax of `merge`

weralabaj:
@SzymonPobiega I wonder if actually locking level is an issue here really. I mean the behavior was as desired, timeout subscriptions were processed, so I think ensuring proper logging level (info probably is too high for internal stuff that shouldn't concern users) and using dedicated retries instead of FLR should be a satisfactory solution. What do you think?

SzymonPobiega:
Yeah. Also, I was not able to reproduce this behavior in my tests :(

weralabaj:
So maybe we are a bit overzealous with the changes? Happy to get more opinions, but I see in the related case you remarked that logged error messages that are below "error" level are automatically resolved and according to our philosophy should not be a problem. And that makes sense IMO, especially in this case where potential deadlocks occur only at endpoint startup.

weralabaj:
@Particular/sqlpersistence-maintainers your input in the discussion would be appreciated.

MarcinHoppe:
This change has been beneficial for NH persistence in the past so I'm 👍 to include it here as well.

SzymonPobiega:
I'll rebase this.

src/SqlPersistence/Subscription/SubscriptionPersister.cs
```diff
@@ -32,32 +32,61 @@ public SubscriptionPersister(Func<DbConnection> connectionBuilder, string tableP
...
+        var attempts = 0;
+        while (true)
+        {
+            try
+            {
+                await action().ConfigureAwait(false);
+                return;
+            }
+            catch (Exception)
+            {
```
SzymonPobiega: Yep. Done.

SzymonPobiega:
@weralabaj you can press the big green button when it finishes building :-)

Issue #112, opened 2017-06-14
Subscription store does not accept subscribe requests issued from NServiceBus 5 and below


SzymonPobiega:
### Who's affected

Users who have mixed-version environments with some endpoints running NServiceBus 5 and some running NServiceBus 6

### Symptoms

Subscribe requests issued by NServiceBus 5 endpoints are rejected by NServiceBus 6 endpoints and eventually end up in the `error` queue. This, if not detected, can cause message loss because the subscription is not created. In case where there is only one subscriber and it failed to subscribe, the messages published are lost.

### Backported to

- [x] 1.0.4: #114 
- [x] 2.0.0: #111

MarcinHoppe:
Not all supported versions have been patched yet, reopening.

SzymonPobiega:
@MarcinHoppe the reason I initially closed it was to be able to generate release notes. Working with multi-version patches and release notes generator is a bit painful.

MarcinHoppe:
Resolved for all supported versions, closing.

PR #113, opened 2017-06-14
Subscription store does not accept subscribe requests issued from NServiceBus 5 and below

DavidBoike:


MarcinHoppe:
Resolves #112

PR #121, opened 2017-06-26
WIP: NuGetPackager Version="0.6.4"

SimonCropp:
waiting for NuGetPackager to be deployed


SzymonPobiega:
@SimonCropp I actually added code so that the existing `<projectUrl/>` will be handled correctly so no need to remove it for other nuspecs.

PR #126, opened 2017-07-07
Add ability to toggle script generation

SimonCropp:
Adds the ability to toggle script creation for specific storage types.

Unlocks the ability to mix this persister with other persisters

Note: Moving the storage types (`s.EnableFeature<StorageType.Outbox>();`) into the PersistenceDefinition means those settings can be checked in the `StorageSessionFeature`.


SimonCropp:
@SzymonPobiega thoughts?

src/SqlPersistence/SynchronizedStorage/StorageSessionFeature.cs
```diff
@@ -1,14 +1,23 @@
...
 class StorageSessionFeature : Feature
 {
 
     protected override void Setup(FeatureConfigurationContext context)
     {
-        var connectionBuilder = context.Settings.GetConnectionBuilder();
+        var settings = context.Settings;
+        var isSagasEnabled = settings.GetFeatureEnabled<StorageType.Sagas>();
+        var isOutboxEnabled = settings.GetFeatureEnabled<StorageType.Outbox>();
+        if (!isSagasEnabled || !isOutboxEnabled)
```
SzymonPobiega: Why do we require this?

SzymonPobiega:
@SimonCropp Can you elaborate, in the description, what's the purpose of changes in `SqlPersistence.cs` ? Is it still a `WIP`?

src/SqlPersistence/SynchronizedStorage/StorageSessionFeature.cs
SimonCropp: to prevent the runtime case exception when mixing persistences

SzymonPobiega: Hoa about a case where somebody enables the persistence explicitly for sagas but not for outbox (because he does not use the outbox)?

SimonCropp: good point. i will add that scenario

SimonCropp:
WIP removed. updated the PR description

SimonCropp:
@SzymonPobiega how does it look now?

SzymonPobiega:
@SimonCropp looks good. My only concern is what would happen if somebody uses storage session as described here: https://docs.particular.net/samples/sqltransport-sqlpersistence/#receiver-project-accessing-the-ambient-database-details but does not have a saga and outbox is disabled. Would it still work? I think the saga and outbox persistence concerns would not get activate so the session would also be disabled.

Otherwise we would have to tell the users to use the transport connection/transaction in such a case.

SimonCropp:
Thats why i was asking previously if the sql transport should also expose some kind of session

SzymonPobiega:
@SimonCropp My personal opinion is that is should but maybe we should discuss this in a bigger group. My argument against that is based on what was happening prior to 6. You had two distinct API, one in the transport and one in NH persistence for getting connection and transaction. Depending on the fact if there was a saga in the scanned dlls you would have to use one or the other. The fragility of this is IMO an argument to always tell users to use persistence and not go directly to the transport.

And as I said, ideally we would have saga persistence decoupled with its own message deduplication. If we did then the saga handlers could happily be called at-least-once and the user would have to either use transport connection or the outbox connection. Enabling (or not) outbox is much more stable decision that having (or not) a saga in scanned dlls.

SimonCropp:
happy to discuss that in a broader group.

But i still think this PR is valid. eg if u r only using the sql persistence for timeouts or subscriptions then it makes no sense for the sql persistence be providing an abstraction over the sql transport transaction/connections.

@SzymonPobiega so do u have outstanding objections to this PR as is? or should we merge and deploy to unlock @danielmarbach use case?

SzymonPobiega:
@SimonCropp apart from philosophical discussion if we should or should not expose the session, the fact is we used to expose it and now we're breaking this behavior which I think requires a major bump. I don't think we said in the doco that the synchronized session is available only when you use outbox or sagas and hard for me to imagine we can say this behavior change is a bug fix.

SimonCropp:
@SzymonPobiega how about we go with this as a major. then if anyone wants the sql persistence to wrap when only using timeouts/subscriptions etc we can add it as an opt in in a minor?

SzymonPobiega:
@SimonCropp sounds like a good idea. 👍

DavidBoike:
So heads up that multi-catalog support also looks like it would be a breaking change requiring a major. Might want to combine those to minimize majors as well as argue for prioritization of multi-catalog.

SimonCropp:
@DavidBoike timeline for multi-catalog?

DavidBoike:
It's Next in Line, Needs Taskforce, so I think not long.

PR #134, opened 2017-07-22
[WIP] Use project system to build packages

bording:
This is an attempt to move to using the new project system's support for producing NuGet packages directly, which is a nice thing to have when moving to NSB 7 / multi-targeting to netcoreapp2.0.

I've marked this [WIP] for now because there's a blocker in producing Release builds: both ScriptBuilder and ScriptBuilderTask want to merge Mono.Cecil, and the sequencing of this now means that the already merged ScriptBuilder  assembly is used in ScriptBuilderTask's compile step.

Looking at the code, I think it makes sense to move the relevant code out of ScriptBuilderTask and into ScriptBuilder, which would remove the reference to Mono.Cecil in ScriptBuilderTask and solve this issue.

I wanted to run this by everyone before actually making those changes though!


SimonCropp:
i think it is more complicated than that. the test libs also use cecil. the problem is that the ilmerging has to happen as part of the packaging and not effect the bin files. so i think we have two options

 1. work out how to hack the packaging merge before pack but not effect the refs of the other projects

or

 2. dont merge and ref cecil as a lib that is checked in. and hack the name of that file to not conflict with other copies of cecil in the same appdomain

bording:
> i think it is more complicated than that. the test libs also use cecil. 

Fair enough. I still think it makes sense to remove the direct usage in ScriptBuilderTask and do it all in ScriptBuilder instead.

I had originally put the merging target where I did so it would execute inside the inner build, removing the need to figure what TFMs were in the project.

I've got an idea for how I can get that same TFM-specific aspect during pack: make the target run during pack with `TargetsForTfmSpecificContentInPackage` like I'm doing for including the PDBs.  That should get us what we need without breaking referencing builds.

bording:
OK, with this latest commit, the merging is now invoked as part of the Pack target. I did have to remove the `GeneratePackageOnBuild` property, because that invokes Pack for each project separately as part of the build process, causing the same issue.

Now, Pack needs to be invoked manually, which allows for the correct sequencing. That does mean the CI build step needs to be modified to include Pack in the list of targets it invokes, but once that it done, this should work and create packages.

SimonCropp:
replaced by https://github.com/Particular/NServiceBus.Persistence.Sql/pull/135

PR #135, opened 2017-07-30
Use new packaging

SimonCropp:
so this is based on https://github.com/Particular/NServiceBus.Persistence.Sql/pull/134/commits/427430f10336f5ded2abcc891735fa1836a77162

it moves the nuget creation over to full native ms tooling. packages are now created as part of a build with no other tooling necessary. note this is the reason we are currently on the net core TC agent

and most of the credit goes to @bording for getting this working

@bording i also moved the code around to avoid the need to calling pack on the build server


SimonCropp:
@bording one thing i was not sure about was your addition of `<CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>`?  what does that do? can u point to some doco?

SimonCropp:
@Particular/sqlpersistence-maintainers this one is ready for a review

src/Integration/Shared/EndpointStarter.cs
```diff
@@ -22,38 +22,32 @@ public static class EndpointStarter
         var endpoint = await Endpoint.Start(endpointConfiguration)
             .ConfigureAwait(false);
         Console.WriteLine("Press any key to exit");
-        try
+        var myEvent = new MyEvent
```
weralabaj: How's that change related?

weralabaj:
@bording Looks good to me, but since I don't have the context regarding those changes I think you should also have a look.

src/Integration/Shared/EndpointStarter.cs
SimonCropp: TBH it isnt really. i just noticed it when trouble shooting a build issue

bording:
@SimonCropp `CopyLocalLockFileAssemblies` is how you set `CopyLocal` to `true` for all assemblies referenced in the lock file (project.asset.json), so you get a copy of them in the output folder instead of them just being referenced directly from the global NuGet package folder. This is what allows the ILRepack target to be much simpler, since all of the assemblies are right where you need them to be.

That means you should only need this set on the ScriptBuilderTask project now, since that's the only project that's still merging dependencies.

As far as docs on it are concerned, I learned about it from various GitHub issues. I haven't seen it mentioned anywhere else yet.

bording:
There are some changes here compared to my PR that have some impact that need to be considered, so I'll leave some comments.

src/nuget.config
```diff
@@ -8,7 +8,6 @@
   <packageSources>
     <add key="nuget.org" value="https://www.nuget.org/api/v2/" />
     <add key="particular myget.org" value="https://www.myget.org/F/particular/" />
-    <add key="Local Integration Source" value="..\nugets" />
```
bording: Why did you remove this? I'm not sure how the Integration solution will be able to build properly without this, but maybe I just don't understand how it's used?

src/MsSqlAcceptanceTests/MsSqlAcceptanceTests.csproj
```diff
@@ -7,7 +7,6 @@
     <AssemblyOriginatorKeyFile>$(SolutionDir)Test.snk</AssemblyOriginatorKeyFile>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
```
bording: This isn't redundant. As soon as you need to test on .NET Core, you'll need this, and you'll also need NUnit3TestAdapter.

src/Directory.Build.props
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
```
bording: Why did you change this to a props file? This was intentionally a targets file because of the order of evaluation. Props files get evaluated before items in the project file, and targets files get evaluated after the project file. There items in here that need the project file to get a chance to add values first.

src/Directory.Build.props
```diff
@@ -0,0 +1,25 @@
...
+<Project>
+
+  <PropertyGroup>
```
bording: The properties that I had in here were intentionally focused specifically on NuGet-related things that could be pulled into a separate package and included into every package. If you want to have extra stuff in here, it would be nice to at least put it into a separate PropertyGroup to make this easier to migrate.

src/Directory.Build.props
```diff
@@ -0,0 +1,25 @@
...
+    <Authors>NServiceBus Ltd</Authors>
+    <Company>NServiceBus Ltd</Company>
+    <PackageLicenseUrl>http://particular.net/LicenseAgreement</PackageLicenseUrl>
+    <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
+    <Copyright>Copyright 2010-$([System.DateTime]::UtcNow.ToString(yyyy)) NServiceBus. All rights reserved</Copyright>
+    <PackageTags>nservicebus servicebus cqrs publish subscribe</PackageTags>
+    <PackageIconUrl>http://s3.amazonaws.com/nuget.images/NServiceBus_32.png</PackageIconUrl>
+    <PackageProjectUrl Condition="'$(PackageProjectUrl)' == ''">https://docs.particular.net/nuget/$(PackageId)</PackageProjectUrl>
```
bording: This condition is here to allow you to override the url in the project file, with the idea of this being generic and included in our upcoming new NuGet package.

src/Directory.Build.props
```diff
@@ -0,0 +1,25 @@
...
+    <PackageOutputPath>..\..\nugets</PackageOutputPath>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <TargetsForTfmSpecificContentInPackage>$(TargetsForTfmSpecificContentInPackage);IncludePDBsInPackage</TargetsForTfmSpecificContentInPackage>
+    <PackageProjectUrl>https://docs.particular.net/nuget/nservicebus.persistence.sql</PackageProjectUrl>
```
bording: The condition above was why I had this defined in the two projects that needed to override the default value.

src/Integration/Directory.Build.props
```diff
@@ -0,0 +1 @@
+<Project><PropertyGroup><IntegrationVersion>3.0.0-fixpackaging0001</IntegrationVersion></PropertyGroup></Project>
```
bording: I've now released that for this work the way I intended it, we probably don't want to check this in, since it will always be changing.

src/Integration/Shared/Shared.csproj
```diff
@@ -1,16 +1,12 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net462</TargetFramework>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
SimonCropp: why do we need a Directory.Build.props in the Integration dir?

src/ScriptBuilder/ScriptBuilder.csproj
```diff
@@ -1,50 +1,36 @@
...
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>$(SolutionDir)NServiceBus.snk</AssemblyOriginatorKeyFile>
-    <DebugType>full</DebugType>
-    <DebugSymbols>True</DebugSymbols>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <OutputPath>..\..\binaries\</OutputPath>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <NoWarn>CS1591</NoWarn>
-    <UpdateAssemblyInfo>true</UpdateAssemblyInfo>
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
```
SimonCropp: done

src/ScriptBuilder/ScriptBuilder.csproj
```diff
@@ -1,50 +1,36 @@
...
-    <Compile Include="..\Guard.cs" />
-    <Compile Include="..\CommonAssemblyInfo.cs" />
-    <None Remove="Outbox\*.sql" />
-    <None Remove="Subscription\*.sql" />
-    <None Remove="Timeout\*.sql" />
-    <EmbeddedResource Include="Outbox\*.sql" />
-    <EmbeddedResource Include="Subscription\*.sql" />
-    <EmbeddedResource Include="Timeout\*.sql" />
+    <PackageReference Include="GitVersionTask" Version="4.0.0-beta0012" PrivateAssets="All" />
+    <PackageReference Include="ILRepack" Version="2.0.13" PrivateAssets="All" />
```
SimonCropp: done

src/Directory.Build.props
SimonCropp: but these nugets don't have different doco urls

src/Integration/Directory.Build.props
SimonCropp: why do we even need this file?

SzymonPobiega:
@SimonCropp @bording can you extract some kind of `how-to` from this PR so that it is easier to migrate other repos? Then we can deprecate the NGP.

bording:
@SzymonPobiega Now that we've been able to migrate a few repos and validated that it will work for even some of the more complex packaging scenarios, the idea was the take the parts that that can be shared and add them via a targets file brought in via a new package. That combined with a bit of a `how-to` guide should make migrating pretty easy.

Opening an issue in platdev about this has been on my todo list, but I haven't gotten to it yet.

PR #142, opened 2017-08-04
sql dialect class

weralabaj:
Connects to https://github.com/Particular/PlatformDevelopment/issues/1382.


weralabaj:
With this approach, would it make sense to move connection and table prefix settings also to the new classes rather than operate directly on persistence settings?

i.e. have `dialectSettings.ConnectionBuilder(MsSqlConnectionBuilder.Build)` and `dialectSettings.TablePrefix($"{tablePrefix}_");`

That would give result in a more consistent API, but then is change that is not really necessary (more to change during upgrade).

Thoughts?

DavidBoike:
I think if we're going to make the breaking change we should go for the more consistent API.

Also in Oracle there are some elements like table prefix where, because of Oracle naming constraints, we ignore. So advantage of making that method dialect-specific beyond just an API that's "more consistent."

Let me know how I can help. I don't want to step on your toes while you've got it marked as WIP.

SzymonPobiega:
So the only visible change so far is that the `Schema` configuration will be only available when I chose the SqlServer dialect, right?

This will pave the way for adding `Catalog` configuration etc.

weralabaj:
@SzymonPobiega correct, next step would be to similarly limit other settings which make sense (if you agree more breaking changes are justified)

weralabaj:
@DavidBoike I didn't manage to get to it today, trying to complete a workshop offer before Yifat's vacation and some other stuff. If you have a moment, the next step would be to modify API to change other configuration methods to also use the new types.
Also, @SzymonPobiega suggested that it would be a better idea to pass a class instance, rather than type and use polymorphism to handle the differences between specific implementations (if I got the intention right). I started in that direction, but there were many changes at once and had issues getting it compile, so abandoned it for a moment, but I agree that should be a better approach ultimately.

DavidBoike:
Pushed a commit that changes the Type-based switching to a polymorphism based one, and storing the settings in the SqlDialect instance (in the settings) instead of in disparate settings. So far, only Schema is handled in this way, but I think this will work fairly well as we can, one-by-one, remove parameters that get tossed around (tablePrefix, etc.) and contain them within just the SqlDialect instance, eventually to the point where it's the dialect instance that is the single source of truth for how to construct a table name, parameter name, etc.

Would love some feedback @SzymonPobiega @SimonCropp.

Didn't try the acceptance tests yet (just due to volume and time) but all the other tests pass. Also not sure I captured all renamings of sqlVariant to sqlDialect - figured a big search and replace would be in order at some point down the line.

DavidBoike:
Committed the last few details, all tests pass on build server (build agents using .NET Core 2.0 final, and as a result, the MSSQL tests now pass), and I think this is in a reviewable state.

src/SqlPersistence.Tests/Saga/SagaPersisterTests.cs
```diff
@@ -47,8 +47,7 @@ SagaPersister SetUp(string endpointName)
             readerCreator: reader => new JsonTextReader(reader),
             writerCreator: writer => new JsonTextWriter(writer),
             tablePrefix: $"{endpointName}_",
-            schema: schema,
-            sqlVariant: runtimeSqlVariant,
+            sqlDialect: runtimeSqlVariant,
```
SimonCropp: shouldnt we move all "Variant" over to "Dialect? so future people dont need to mind map on variable/param/method names?

src/SqlPersistence.Tests/SqlVariantConverter.cs
```diff
@@ -1,35 +1,48 @@
 using System;
-using NServiceBus.Persistence.Sql;
+using NServiceBus;
 using NServiceBus.Persistence.Sql.ScriptBuilder;
 
 public static class SqlVariantConverter
 {
-    public static BuildSqlVariant Convert(this SqlVariant sqlVariant)
+    public static BuildSqlVariant Convert(this SqlDialect sqlDialect)
```
SimonCropp: any reason why BuildSqlVariant  wasnt renamed?

src/SqlPersistence/Config/SqlDialect.cs
```diff
@@ -0,0 +1,48 @@
+namespace NServiceBus
+{
+    /// <summary>
+    /// Allows for configuring which database engine to target. Used by <see cref="SqlPersistenceConfig.SqlDialect{T}"/>.
+    /// </summary>
+    public class SqlDialect
```
SimonCropp: should this have a protected constructor?

src/SqlPersistence/Config/SqlPersistenceConfig_Schema.cs
```diff
@@ -1,38 +1,30 @@
...
-
     //TODO: throw for schema in mysql
     public static partial class SqlPersistenceConfig
     {
 
         /// <summary>
-        /// Configures the database schema to be used.
+        /// Obsolete: Use 'persistence.UseSchema&lt;SqlDialect.DialectType&gt;().Schema("schema_name")' instead. Will be removed in version 4.0.0.
         /// </summary>
+        [Obsolete("Use 'persistence.UseSchema<SqlDialect.DialectType>().Schema(\"schema_name\")' instead. Will be removed in version 4.0.0.", true)]
```
SimonCropp: @DavidBoike yeah as u pointed out u can re-ad fody here if you want to

SimonCropp:
some very minor feedback. the rest looks good. 

once others have reviewed and this is merged on develop, perhaps add the changes to the upgrade guide, snippet, and samples and lets see if the resultant real usage looks as good as this PR?

DavidBoike:
Per @SimonCropp's suggestions:

* Renamed all instances of "variant" as a concept to "dialect"
* Added the protected constructor for SqlDialect
* Added Obsolete.Fody back in - now my confidence that the obsoletes are all good and consistent is much higher. Fixed one I had the absolute wrong method name in the replacement suggestion.

The renames in particular made the size of this PR a lot bigger though. I recommend reviewing ranges of commits - everything through August 14 (equivalent of what Simon reviewed) and then perhaps commit-by-commit after that, plus the API Approvals at the "all commits" level.

Agree wholeheartedly with Simon's suggestion to update doco immediately after this PR is merged.

@SzymonPobiega I hope you can review this as one of your first things back tomorrow?

src/SqlPersistence/Config/SqlDialect.cs
SzymonPobiega: 👍 for `abstract` + protected `ctor`

PR #145, opened 2017-08-16
Replaced all the ifs based on dialect type with abstract/virtual methods on the dialect itself.

SzymonPobiega:
The `SqlDialect` implementations are split so that there is one partial per persistence type / dialect pair so that the code that is used together sits together. One can also look up all the SQL code for a given dialect by using `cttl+T` and typing e.g. `_Oracle`.


src/SqlPersistence/Config/SqlDialect_Oracle.cs
```diff
@@ -0,0 +1,56 @@
...
+                {
+                    parameter.Value = value;
+                }
+            }
+
+            internal override CommandWrapper CreateCommand(DbConnection connection)
+            {
+                var command = connection.CreateCommand();
+
+                var bindByNameProperty = command.GetType().GetProperty("BindByName");
```
DavidBoike: Did we have this reflection optimized in any way before? Should we do so now?

src/SqlPersistence/SqlHelpers.cs
```diff
@@ -1,58 +1,40 @@
-﻿using System.Collections.Generic;
-using System.Data.Common;
+﻿using System.Data.Common;
 using System.Threading.Tasks;
 
 static class SqlHelpers
 {
+    //public static async Task ExecuteTableCommand(this DbConnection connection, DbTransaction transaction, string script, string tablePrefix, string schema)
```
DavidBoike: Why keep an internal static class with its members completely commented out?

SzymonPobiega: @DavidBoike I missed this one. I removed that class on `develop`.

src/SqlPersistence/Config/SqlDialect_Oracle.cs
SzymonPobiega: Nope, I just copied the code there. I tried to optimize it here https://github.com/Particular/NServiceBus.Persistence.Sql/pull/147

PR #147, opened 2017-08-17
Optimize BindByName access for Oracle

SzymonPobiega:


src/SqlPersistence/Config/SqlDialect_Oracle.cs
```diff
@@ -12,6 +13,8 @@ public abstract partial class SqlDialect
         /// </summary>
         public partial class Oracle : SqlDialect
         {
+            volatile PropertyInfo bindByNameProperty;
```
SimonCropp: i though volatile was bad for perf on the hot path https://blogs.msdn.microsoft.com/brada/2004/05/13/volatile-and-memorybarrier/

SzymonPobiega: It is not ideal but I think better than `.GetType().GetMethod()`. I put volatile here to ensure that eventually the value can be read by other threads. To be honest I don't think it is strictly necessary.

DavidBoike: And it can't be optimized into a static constructor or something like that because it has to wait for an OracleCommand object to pass through.

I'm not an expert on the volatile keyword by any means. Is something like this better, attempting to ensure that the instance of the PropertyInfo you're trying to deal with is always a local variable?

```
var localProperty = bindByNameProperty;
if(localProperty == null)
{
    localProperty = command.GetType().GetProperty("BindByName");
    bindByNameProperty = localProperty;
}
localProperty.SetValue(command, true);
```

SzymonPobiega: Let's ask @Scooletz . He's The Man!

PR #165, opened 2017-08-31
Do not run installers when persistence is not configured

### LINKED ISSUE #109: SQL Persistence endpoints cannot be hosted in a same process as endpoints that use other persistences
The `Installer` class is loaded to all endpoints because of the assembly scanning. In NHibernate there is a work-around that only triggers the schema creation if the corresponding feature is indeed activated. In the SQL persistence is is executed regardless. See https://github.com/Particular/NServiceBus.NHibernate/blob/develop/src/NServiceBus.NHibernate/TimeoutPersisters/NHibernateTimeoutStorage.cs#L25 for details.

SzymonPobiega:
Fixes #109 

The change introduces an explicit`InstallerSettings` class that is used to pass the parameters to the installers as well as to control if the installer should run. Installer classes are registered automatically by the core so they are executed even if the persistence dll is only present in `bin` but not actually configured properly.

With this change is SQL persistence is referenced but not configured via `EndpointConfiguration` the installer won't run. It will still be registered in the container by the core, but `InstallerSettings` object will not be present in the settings bag because the corresponding feature did not run. The feature is only enabled when user configures the persistence in `EndpointConfiguration`.

Question to reviewers: how do I test it? I have no idea how to test the whole installer machinery.

### DIFF
--- a/src/SqlPersistence/Installer.cs
+++ b/src/SqlPersistence/Installer.cs
@@ -4,53 +4,47 @@ using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using NServiceBus;
+using NServiceBus.Features;
 using NServiceBus.Installation;
 using NServiceBus.Logging;
 using NServiceBus.Settings;
 
 class Installer : INeedToInstallSomething
 {
+    InstallerSettings installerSettings;
     ReadOnlySettings settings;
     static ILog log = LogManager.GetLogger<Installer>();
 
     public Installer(ReadOnlySettings settings)
     {
         this.settings = settings;
+        installerSettings = settings.GetOrDefault<InstallerSettings>();
     }
 
     public async Task Install(string identity)
     {
-        //If neither of the features is configured then do not even attempt to validate config
-        if (!settings.ShouldInstall<SqlOutboxFeature>()
-            && !settings.ShouldInstall<SqlSubscriptionFeature>()
-            && !settings.ShouldInstall<SqlTimeoutFeature>()
-            && !settings.ShouldInstall<SqlSagaFeature>())
+        if (installerSettings == null || installerSettings.Disabled)
         {
             return;
         }
 
-        var connectionBuilder = settings.GetConnectionBuilder();
-        var sqlDialect = settings.GetSqlDialect();
-        var scriptDirectory = ScriptLocation.FindScriptDirectory(sqlDialect);
-        var tablePrefix = settings.GetTablePrefix();
+        installerSettings.Dialect.ValidateTablePrefix(installerSettings.TablePrefix);
 
-        sqlDialect.ValidateTablePrefix(tablePrefix);
-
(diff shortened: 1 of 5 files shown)


weralabaj:
Hm... the only option of testing that comes to my mind is a smoke test with some sample, also somehow we must have noticed this is an issue, how does that manifest? Or was that just a code analysis? :)

src/SqlPersistence/Installer.cs
```diff
@@ -4,53 +4,47 @@
...
 
             transaction.Commit();
         }
     }
-
+    
     Task InstallOutbox(string scriptDirectory, DbConnection connection, DbTransaction transaction, string tablePrefix, SqlDialect sqlDialect)
     {
-        if (!settings.ShouldInstall<SqlOutboxFeature>())
+        if (!settings.IsFeatureActive(typeof(SqlOutboxFeature)))
```
weralabaj: Just looking at this piece of code, without analyzing context in depth... It's a bit confusing, as you may have a feature active but don't want to execute installers (e.g. in prod). Maybe it's just a matter of terminology, but I'll definitely stop here for a moment in the future to think.

weralabaj: Just looking at this piece of code, without analyzing context in depth... It's a bit confusing, as you may have a feature active but don't want to execute installers (e.g. in prod). Maybe it's just a matter of naming, but I'll definitely stop here for a moment in the future to think.

weralabaj:
> It will be missing because it is created by defaults in InstallerFeature which is only enabled when the persistence is enabled.

Maybe it's a mental shortcut but how does that solve the issue of having two persistences in one process? And enabling persistence is not a sufficient condition to have installers run... So I don't understand :)

src/SqlPersistence/Installer.cs
SzymonPobiega: The "installers disabled for prod" condition is handled in the calling method. The check here is only to install only the bits that are required given the feature set configured in the endpoint e.g. if there is no sagas, do not run saga installers.

SimonCropp:
@SzymonPobiega is it possible to have an integration test that verifies this?

SzymonPobiega:
I've tests it with an integration sample. When switched to `InMemoryPersistence` no tables were created in the database. When switched back to SQL but disabled the outbox, all tables but the outbox one were created. When enabled the outbox back, all tables were created. This should cover all the angles.

SimonCropp:
but no way to add a test to the codebase?

SzymonPobiega:
I could add a test like the ones that verify saga/user data consistency but I think this is not worth the effort. The test itself would be quite simply but checking the result and resetting the state would be more tricky (need to remove the database before each run to verify it tables are/are not created).

weralabaj:
IMO it sounds like something to smoke test, automation not worth the ROI, so I agree with Szymon.

PR #169, opened 2017-09-05
WIP: faster install via better async

SimonCropp:


SimonCropp:
@Particular/sqlpersistence-maintainers thoughts on detecting MARS and executing installation scripts in parallel? and on another note should we support diff connection settings for the installation script execution

SzymonPobiega:
I don't have enough knowledge to have an opinion. I always thought that MARS was designed for scenarios where you have a parent query and a child query and for each result of the master query you want to issue a child query on the same connection.

SimonCropp:
@SzymonPobiega yeah i dont have much experience with MARS either. i only when down that path when  attempting to do whenall on the scripts i got an error saying i needed to enable MARS.

SimonCropp:
TBH thinking this is not worth the complexity.

DavidBoike:
I think MARS is a fairly archaic, not well-known option, so I'd agree it's not worth the complexity given that the user would have to know about the option, use it, and then may not even notice the speed difference. Especially since we don't intend users to use the installers in production, it would seem fairly odd to recommend the MARS setting for an experience only at development time, which may have other ramifications for them if they carried it through to prod.

Issue #157, opened 2017-08-23
Buffer pooling for saga operations


Scooletz:
Currently, the saga persister always creates a string containing saga data, by calling `ToJson`. Some dialects, like SQLServer provides capabilities of passing [a size of the parameter](https://msdn.microsoft.com/en-us/library/system.data.sqlclient.sqlparameter.size(v=vs.110).aspx), which can be useful, especially, when combined with some buffer management.

Have you @Particular/sqlpersistence-maintainers considered this optimization? Would it be something you'd be interested into?

https://github.com/Particular/NServiceBus.Persistence.Sql/blob/3b234e1da4652759b982c1c1f6ec39236f62c7a8/src/SqlPersistence/Saga/SagaPersister_Update.cs#L27

SzymonPobiega:
@Scooletz sounds good. I don't think we considered it.

MarcinHoppe:
To add what @SzymonPobiega said, we wouldn't mind a PR 😉.

Scooletz:
Solved by #175

PR #175, opened 2017-09-18
Memory optimized saga serialization

Scooletz:
This PR introduces a memory optimizee saga serialization for Update and Save operations, capable of using custom pooled `TextWriter` objects.

### Implementation details

This PR introduces a pooled `TextWriter` that enables efficient passing of serialized sagas to `MsSqlServer` dialect using `.Offset` and `.Size` of `SqlParameter`, effectively enabling working with `ArraySegment<byte>` and `ArraySegment<char>` without additional allocations. 

The memory efficiency of this PR is a result of removal of `StringBuilder.ToString` that allocates a string every time it's called. `StringBuilder` uses internally some kind of pooling, but still, the result string needs to be allocated. By writing directly to a custom `TextWriter` and accessing its internal storage `ArraySegment<char>` we can use the mechanism described above and skip the final allocation.

One could argue about a possible leakage of this pool, but as both, a concurrency of an endpoint and the number of sagas handling a specific message type is limited, I'd say it's safe to leave the pool as unbounded and growing when needed.

### Bechmarking

Benchmarking was done using a dummy loop over sagas to be stored, that contained following data:

```c#
var data = new SagaData
{
    StringProperty = "Some bigger test",
    GuidProperty = Guid.NewGuid(),
    OriginalMessageId = Guid.NewGuid().ToString(),
    Originator = "originator",
    IntProperty = 5,
    Id = Guid.NewGuid()
};
```

Results:

Number of serializations: 10 mln

| Dialect   | Time  |
|---|---|
| MsSqlServer | 00:00:18.7940289 |
| MySql | 00:00:21.0636780

Observations:
- the bigger saga, the bigger gain
- benchmarkdotnet could be useful
- as always time spending on allocations adds up

### TODO:
- [x] `CharArrayTextWriter` requires testing
- [x] a potential benchmark comparing JSONization with these approaches

Addresses #157

### DIFF
--- a/src/SqlPersistence.Tests/CharArrayTextWriterPerformanceTests.cs
+++ b/src/SqlPersistence.Tests/CharArrayTextWriterPerformanceTests.cs
@@ -1,0 +1,107 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using NServiceBus;
+using NServiceBus.Persistence.Sql;
+using NServiceBus.Sagas;
+using NUnit.Framework;
+using NUnit.Framework.Interfaces;
+
+[Explicit("This is a long running performance test")]
+public class CharArrayTextWriterPerformanceTests
+{
+    [TestCaseSource(nameof(Dialects))]
+    public void PerfTest(SqlDialect dialect)
+    {
+        var sagaInfo = BuildSagaInfo<S,S.SagaData>(dialect);
+        var data = new S.SagaData
+        {
+            StringProperty = "Some bigger test",
+            GuidProperty = Guid.NewGuid(),
+            OriginalMessageId = Guid.NewGuid().ToString(),
+            Originator = "originator",
+            IntProperty = 5,
+            Id = Guid.NewGuid()
+        };
+
+        // warm up
+        using (var cmd = new CommandWrapper(null, dialect))
+        {
+            Console.WriteLine(dialect.BuildSagaData(cmd, sagaInfo, data));
+        }
+
+        var sw = Stopwatch.StartNew();
+        for (var i = 0; i < 1000000; i++)
+        {
+            using (var cmd = new CommandWrapper(null, dialect))
+            {
+                dialect.BuildSagaData(cmd, sagaInfo, data);
+            }
+        }
+        Console.WriteLine($"Test took: '{sw.Elapsed}'");
+    }
+
+    public class S :
+        SqlSaga<S.SagaData>,
+        IAmStartedByMessages<AMessage>
+    {
+        public class SagaData : ContainSagaData
+        {
+            public string StringProperty { get; set; }
+            public int IntProperty { get; set; }
+            public Guid GuidProperty { get; set; }
+        }
+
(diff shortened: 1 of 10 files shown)


Scooletz:
Ping @Particular/sqlpersistence-maintainers I'd like to see your opinion on this approach before investing time into covering this with additional tests. Should I move forward with it?

DavidBoike:
Looks cool to me.

MarcinHoppe:
Are we hoping for spectacular improvements here? Do we have any sort of indication that saga serialization is the bottleneck? Can be a bottleneck?

Without digging too deep into specific situations, my guess is that most of the time the database I/O would be the bottleneck.

Scooletz:
@MarcinHoppe I think that depends on the spectacularness definition.

I don't follow the IO-argument. IO is always the slowest, still we care about NSB that is fast and robust. If we compare anything to IO, then relatively we should not care about it.

Anyway, maintainers, you've got numbers, and this is ready for a review.

src/SqlPersistence.Tests/CharArrayTextWriterPerformanceTests.cs
```diff
@@ -0,0 +1,107 @@
...
+            return Task.FromResult(0);
+        }
+    }
+
+    public class AMessage : IMessage
+    {
+        public string StringId { get; set; }
+    }
+
+    public static IEnumerable<ITestCaseData> Dialects()
```
MarcinHoppe: Add Oracle.

src/SqlPersistence/Config/SqlDialect_MsSqlServer.cs
```diff
@@ -24,8 +26,20 @@ internal override void AddCreationScriptParameters(DbCommand command)
 
             internal override void FillParameter(DbParameter parameter, string paramName, object value)
             {
-                parameter.ParameterName = paramName;
-                parameter.Value = value;
+                if (value is ArraySegment<char> charSegment)
```
MarcinHoppe: Isn't this required for MySQL and Oracle as well?

src/SqlPersistence.Tests/CharArrayTextWriterPerformanceTests.cs
Scooletz: I picked up MySql only as a comparison. Do you think it's worth to add Oracle?

src/SqlPersistence/Config/SqlDialect_MsSqlServer.cs
Scooletz: `.Offset` and `.Size` are SqlParameter specific and there's no common API for them on the `DbParameter` level. 
I know that with `Npgql` for Postgre you can use this as well, but still, it's db driver specific.

MarcinHoppe: 👍

src/SqlPersistence.Tests/CharArrayTextWriterPerformanceTests.cs
MarcinHoppe: OK, now I understand the driver specific aspect. We can stay with MySQL only.

Scooletz:
😍

PR #178, opened 2017-10-04
Getting netcore on windows tests to run

DavidBoike:


SimonCropp:
@dbelcham @DavidBoike i temporarily disable the inherited build step and added a custom one which does not fail due to missing netcore files on mysql and oracle https://builds.particular.net/admin/editBuildRunners.html?id=buildType:NServiceBusPersistenceSql_32

I did not want to edit the parent step since i dont know what the ramificaions are.

@Particular/sqlpersistence-maintainers merging this since the previous failure was problem with the build server and not this PR

PR #183, opened 2017-10-10
[WIP] Postgresql

weralabaj:
Connected to https://github.com/Particular/PlatformDevelopment/issues/1498.


weralabaj:
- [ ] Add schema support 
- [ ] Fix OutboxPersisterTests.StoreAndCleanup() test

weralabaj:
Design decisions made along the way (should be documented):

### TransactionScope handling 

No support, throw an exception (?)

- [ ] Add a page describing it and/or update existing pages on transactions
- [ ] Update https://docs.particular.net/persistence/sql/accessing-data?version=sqlpersistence_3 (note?)

### Name length limit (by default 63 chars)

No explicit check or shortening convention, Postgres silently substrings names longer than 63 chars. Users may configure their databases to support longer names.

- [ ] Needs to be mentioned [here](https://docs.particular.net/persistence/sql/#supported-sql-implementations-supported-name-lengths)

### ...

PR #187, opened 2017-10-12
Quoting aroud ILRepack

Scooletz:
If a user directory path contains spaces, path to the ILRepack should be quoted to work.

This is what we were hit with @seanfarmar when applying ILRepack across some Metrics packages


SzymonPobiega:
❤️

PR #188, opened 2017-10-18
Using temp schema pg_temp for functions

SzymonPobiega:


src/ScriptBuilder/Outbox/Create_PostgreSql.sql
```diff
@@ -24,4 +24,4 @@
   $body$
   language 'plpgsql';
 
-select create_outbox_table(@tablePrefix);
\ No newline at end of file
+select pg_temp.create_outbox_table(@tablePrefix);
```
weralabaj: > pg_temp.

Making a mental note, we should document using that I think. I'll research more to see if there are any potential problems with this approach. E.g. what if the connection is dropped in the middle of script creation?

src/ScriptBuilder/Saga/Writers/PostgreSqlSagaScriptWriter.cs
```diff
@@ -53,8 +53,9 @@ from information_schema.columns
     public void WriteCreateIndex(CorrelationProperty correlationProperty)
     {
         writer.Write($@"
-createCorrelationIdx = 'create unique index if not exists ""Index_Correlation_{correlationProperty.Name}"" on public.""' || tableNameNonQuoted || '"" using btree (""Correlation_{correlationProperty.Name}"" asc);';
-execute createCorrelationIdx;");
+        script = 'create unique index if not exists ""Index_Correlation_{correlationProperty.Name}"" on public.""' || tableNameNonQuoted || '"" using btree (""Correlation_{correlationProperty.Name}"" asc);';
```
weralabaj: I think the different name here was intentional, because that part will be wrapped in the "create table", but not 100% sure, we need to double check if that won't mess up the "outer" script variable value.

SzymonPobiega: I've checked and the correlation property creation part reused the `script` variable so I figured we can also reuse it here.

PR #194, opened 2017-10-20
[WIP] Attempt to support canceling saga timeouts

SzymonPobiega:
A spike that shows how SQL persistence can implement two forms of timeout canceling:
 * cancel-by-id using user-generated ID
 * cancel-by-type when requesting new timeouts


weralabaj:
Is the intention to release it together with Postgres support or is that independent?

SzymonPobiega:
@weralabaj that's just a spike so I figured it will be easier to base it on postgres since postgres will probably be merged sooner or later and then I'll rebase it on the develop.

SzymonPobiega:
@andreasohlund this is the saga cancellation spike we discussed. Do you think it is worth keeping around? What can we extract from it?

PR #196, opened 2017-10-20
Oracle schema support

lnu:
Schema supported in oracle. I've only tested it with timeout, but I also changed the others in the same way.
I've preserved the original behaviour. If no schema or an empty schema is set, we use the original behaviour. If the schema is set, the table name is generated like this:
```select * from "SCHEMA"."TABLE"```
I did not change the scriptbuilder as I've created all the tables manually.


lnu:
My mistake, I've done the pull request on the wrong branch.

PR #197, opened 2017-10-20
schema support added for Oracle

lnu:
Schema supported for oracle. I've only tested it with timeout, but I've also changed the others in the same way.
I've preserved the original behaviour. If no schema or an empty schema is set, we use the original behaviour. If the schema is set, the table name is generated like this:
select * from "SCHEMA"."TABLE"
I did not change the scriptbuilder as I've created all the tables manually.

### DIFF
--- a/src/SqlPersistence/Outbox/OutboxCommandBuilder.cs
+++ b/src/SqlPersistence/Outbox/OutboxCommandBuilder.cs
@@ -22,7 +22,7 @@ namespace NServiceBus.Persistence.Sql
                     tableName = $"`{tablePrefix}OutboxData`";
                     break;
                 case SqlVariant.Oracle:
-                    tableName = $"{tablePrefix.ToUpper()}OD";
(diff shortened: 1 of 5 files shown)


weralabaj:
Thanks a lot @lnu, we'll review the PR and get back to you.

One question - what version are you using? Is it sufficient to release that change in version 3 (we're currently in beta phase) or you need it in a lower version too?

lnu:
We're currently using version 2. If version 3 is backward compatible with version 2(from a database perspective) it's ok.

MarcinHoppe:
I changed the target branch to `release-2.2` to slate it for the 2.2 release of the transport. I also revived the `2.2.0` milestone.

weralabaj:
@lnu I'll merge your change to the release branch, here https://github.com/Particular/NServiceBus.Persistence.Sql/issues/200 we'll be tracking the new version release. You can follow that issue and I'll ping you once we make a release.

Issue #200, opened 2017-10-23
Support custom schema in Oracle dialect


weralabaj:
Allow to specify custom schema for Oracle dialect.

This change enables to define users/schemas with limited privileges, i.e. without the ability to create/drop tables and configure NServiceBus to run under their credentials.

Related to https://github.com/Particular/NServiceBus.Persistence.Sql/pull/197.

weralabaj:
Implemented and smoke tested in release 2.2.0

PR #202, opened 2017-10-24
Release 2.2

weralabaj:
Related to https://github.com/Particular/NServiceBus.Persistence.Sql/issues/200

## PoA for 

- [x] Fix the [R# inspection error](https://builds.particular.net/viewLog.html?buildId=247926&tab=Inspection&buildTypeId=NServiceBusPersistenceSql_2Insp)
- [x] Merge https://github.com/Particular/NServiceBus.Persistence.Sql/pull/197
- [x] Smoke test all changes in the current release @weralabaj https://github.com/Particular/NServiceBus.Persistence.Sql/milestone/8
- [ ] Document the changes
- [ ] Smoke test other SQL persistence samples
- [ ] Release the version 2.2.0
- [ ] Ping https://github.com/lnu about a release
- [ ] Port fixes to version 3.0
- [ ] Port fix for https://github.com/Particular/NServiceBus.Persistence.Sql/issues/220 to version 1.1

### DIFF
--- a/src/SqlPersistence/Outbox/OutboxCommandBuilder.cs
+++ b/src/SqlPersistence/Outbox/OutboxCommandBuilder.cs
@@ -22,7 +22,7 @@ namespace NServiceBus.Persistence.Sql
                     tableName = $"`{tablePrefix}OutboxData`";
                     break;
                 case SqlVariant.Oracle:
-                    tableName = $"{tablePrefix.ToUpper()}OD";
+                    tableName = string.IsNullOrEmpty(schema) ? $"\"{tablePrefix.ToUpper()}OD\"" : $"\"{schema}\".\"{tablePrefix.ToUpper()}OD\"";
                     break;
                 default:
                     throw new Exception($"Unknown SqlVariant: {sqlVariant}");
@@ -58,7 +58,7 @@ set
 where MessageId = @MessageId";
                 case SqlVariant.Oracle:
                     return $@"
-update ""{tableName}""
+update {tableName}
 set
     Dispatched = 1,
     DispatchedAt = :DispatchedAt,
@@ -86,7 +86,7 @@ where MessageId = @MessageId";
 select
     Dispatched,
     Operations
-from ""{tableName}""
+from {tableName}
 where MessageId = :MessageId";
                 default:
                     throw new Exception($"Unknown SqlVariant: {sqlVariant}");
@@ -114,7 +114,7 @@ values
 )";
                 case SqlVariant.Oracle:
                     return $@"
-insert into ""{tableName}""
+insert into {tableName}
 (
     MessageId,
     Operations,
@@ -149,7 +149,7 @@ where Dispatched = true
 limit @BatchSize";
                 case SqlVariant.Oracle:
                     return $@"
-delete from ""{tableName}""
+delete from {tableName}
 where Dispatched = 1
     and DispatchedAt < :DispatchedBefore
     and rownum <= :BatchSize";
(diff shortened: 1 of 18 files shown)


src/SqlPersistence/Subscription/SubscriptionCommandBuilder.cs
```diff
@@ -29,7 +29,7 @@ public static SubscriptionCommands Build(SqlVariant sqlVariant, string tablePref
                     break;
 
                 case SqlVariant.Oracle:
-                    tableName = $"{tablePrefix.ToUpper()}SS";
+                    tableName = string.IsNullOrEmpty(schema) ? $"\"{tablePrefix.ToUpper()}SS\"" : $"\"{schema}\".\"{tablePrefix.ToUpper()}SS\"";
```
weralabaj: Here we don't call schema.ToUpper but we do in https://github.com/Particular/NServiceBus.Persistence.Sql/pull/202/commits/0ba012f075cc64a60bf9574525faa826f2cf0259#diff-b4d07e6bfe521906a2c801492dd4a453R85, seems like a lack of consistency? not sure if it's ok to always uppercase schemas which may be custom and created without our scripts.

src/SqlPersistence/Outbox/OutboxCommandBuilder.cs
```diff
@@ -22,7 +22,7 @@ public static OutboxCommands Build(string tablePrefix, string schema, SqlVariant
                     tableName = $"`{tablePrefix}OutboxData`";
                     break;
                 case SqlVariant.Oracle:
-                    tableName = $"{tablePrefix.ToUpper()}OD";
+                    tableName = string.IsNullOrEmpty(schema) ? $"\"{tablePrefix.ToUpper()}OD\"" : $"\"{schema}\".\"{tablePrefix.ToUpper()}OD\"";
```
weralabaj: Here we don't call schema.ToUpper but we do in https://github.com/Particular/NServiceBus.Persistence.Sql/pull/202/commits/0ba012f075cc64a60bf9574525faa826f2cf0259#diff-b4d07e6bfe521906a2c801492dd4a453R85, seems like a lack of consistency? not sure if it's ok to always uppercase schemas which may be custom and created without our scripts.

MarcinHoppe:
@weralabaj My memory is failing me. Do you need a hand with testing this?

weralabaj:
To clarify my previous question:

IMO we should **NOT** upper-case schema that is specified by the user, just take it as is and put inside quotes. 

As I understand in all queries/creation scripts, etc. we'll use case-sensitive names (by quoting the strings), so if the schema was created manually (without our new scripts) then queries won't work unless the schema name happens to be upper-cased.

Another approach would be to use case-insensitive comparison for schema name, but that will be inconsistent with what we do for table names, etc.

DavidBoike:
@weralabaj I agree with you. We uppercase table names because they're derived from mixed case endpoint names that play by different rules. When the user specifies a schema they should be specifying it verbatim.

weralabaj:
Smoke tested all changes in the 2.2.0 milestone https://github.com/Particular/NServiceBus.Persistence.Sql/milestone/8

weralabaj:
Smoke test samples with v2.2:
- [ ] https://docs.particular.net/samples/saga/migration/
- [ ] https://docs.particular.net/samples/sql-persistence/transitioning-correlation-ids/
- [ ] https://docs.particular.net/samples/sql-persistence/saga-rename/
- [ ] https://docs.particular.net/samples/saga/sql-sagafinder/
- [ ] https://docs.particular.net/samples/sql-persistence/simple/
- [ ] https://docs.particular.net/samples/outbox/sql/
- [ ] https://docs.particular.net/samples/sqltransport-sqlpersistence/

weralabaj:
Smoke test samples with v1.0.7:
- [ ] https://docs.particular.net/samples/saga/migration/
- [ ] https://docs.particular.net/samples/outbox/sql/
- [ ] https://docs.particular.net/samples/sqltransport-sqlpersistence/
- [x] https://docs.particular.net/samples/performance-counters/ + sql persistence msbuild package

Issue #212, opened 2017-11-01
Custom Saga finding logic for IAmStartedByMessage throws exception


macdonald-keith-vmware:
NSB 6.3.4
Persistence.Sql 2.1.3, MS Sql Server persistence
RabbitMQ 4.3.2

I implemented an IFindSagas<MySagaData>.Using<MyStartCommand> for custom saga finding logic. This works fine if the message being references is for IHandleMessage within the saga. However, if the message is for IAmStartedBy, then several issues occur.

1.  The property used for correlation in the SagaData is not populated. This is easily alleviated by doing the assignment manually. I'm not sure if there can be a resolution for this. 

2. When the IAmStartedBy handler has finished, and the saga is attempted to be persisted, an exception is thrown. This makes it appear that other properties are not being set/managed either.

This can be easily reproduced by using the sample code for saga finding logic on the Particular website (https://docs.particular.net/samples/saga/sql-sagafinder/) and changing the custom finding logic to work for the StartOrder message instead of the CompletePaymentTransaction message.
```
==========================================
System.Exception: Failed to ExecuteNonQuery. CommandText:

insert into [NServiceBus].[mySaga]
(
Id,
Metadata,
Data,
PersistenceVersion,
SagaTypeVersion,
Concurrency,
Correlation_OperationId
)
values
(
@Id,
@Metadata,
@Data,
@PersistenceVersion,
@SagaTypeVersion,
1,
@CorrelationId
) ---> System.Data.SqlClient.SqlException: Must declare the scalar variable "@CorrelationId".
at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
[...]

PR #213, opened 2017-11-02
Validate if correlation property is mapped to a message

SzymonPobiega:
Solves #212


SimonCropp:
Isnt there a work around? Hence doesnt need to be a patch

weralabaj:
Hm... Technically this isn't a valid scenario, so not a bug, so that may not be a patch and can be released with 3.0? Only saying that for this particular user that won't help if it's not on 2.x :)

SzymonPobiega:
Yeah, technically it this does not solve anything. It only adds meaningful validation so I guess we can apply it only to the latest.

weralabaj:
rebased

Issue #15, opened 2017-02-16
Add PostgreSql Support


SimonCropp:

weralabaj:
Consider raising it when we implement AWS support and/or once we have a proper .NET core support.

janovesk:
This would also be nice for Maarten users, btw.

asiemer:
I have a large client that would love to be able to use PostgreSQL (migrating away from Raven) hosted on RDS in AWS.  Thank you.

weralabaj:
Hi @asiemer, since we're talking about moving away Raven, are they interested in Postgres or Marten in particular?

asiemer:
All the spikes I am currently doing are around Marten and Postgres.  Given our timeline I am suggesting a migration to MSSQL/RDS for NSB and app data through Marten/PostgreSQL/RDS.  They like the ease of development and have a significant investment in patterns around Raven.  Need an easy way to swap out Raven and in something else.  Marten/Postgresql solves this it seems on the app dev side.  Once you all release ado/postgres support for NSB...that can follow.  Thoughts?

asiemer:
Is there perhaps an early adopter release of this Postgresql persistence mechanism?  I would love to see how it swaps into their current set up.  Give some feedback.

weralabaj:
@asiemer no, we don't have anything ready at the moment, but knowing there's an interest can affect the priority we give to that idea.

masterpoi:
We are also migrating away from Raven to a Postgresql/Marten setup. Support for postgres in NSB would provide awesome complimentary for us.

weralabaj:
@asiemer @masterpoi We're working on supporting PostgreSQL right now. Here's the PR https://github.com/Particular/NServiceBus.Persistence.Sql/pull/183.

I'm not sure if it's still relevant for you, but if you'd be interested in providing feedback and/or testing it once it's ready, let me know and we'll arrange that.

For now we don't plan on providing special support for Marten though, just standard PostgreSQL.

weralabaj:
@asiemer @masterpoi We're getting close to getting the full Postgres implementation. Would any of you be interested in testing it or discussing in more detail?

Having an external feedback is always appreciated, and can help us catch aspects we didn't consider. If interested, you can let us know here or email me at weronika.labaj at particular.net.

masterpoi:
[...]

PR #219, opened 2017-11-07
throw for bad json.net type metadata handling

SimonCropp:
context https://github.com/Particular/NServiceBus.Persistence.Sql/issues/15#issuecomment-342230295

and after thinking about it for a while i believe we should throw

 * [x] If merged update this doco https://docs.particular.net/persistence/sql/?version=sqlpersistence_3.1#usage-postgresql-json-net-typenamehandling

### DIFF
--- a/src/SqlPersistence/Saga/SqlDialect.cs
+++ b/src/SqlPersistence/Saga/SqlDialect.cs
@@ -24,5 +24,7 @@ namespace NServiceBus
         {
             return sagaInfo.ToJson(sagaData);
         }
+
+        internal abstract void ValidateJsonSettings(Newtonsoft.Json.JsonSerializer jsonSerializer);
     }
 }
--- a/src/SqlPersistence/Saga/SqlDialect.cs
+++ b/src/SqlPersistence/Saga/SqlDialect.cs
@@ -25,6 +25,9 @@ namespace NServiceBus
             return sagaInfo.ToJson(sagaData);
         }
 
-        internal abstract void ValidateJsonSettings(Newtonsoft.Json.JsonSerializer jsonSerializer);
+        internal virtual void ValidateJsonSettings(Newtonsoft.Json.JsonSerializer jsonSerializer)
+        {
+
+        }
     }
 }
--- a/src/SqlPersistence/Saga/SqlDialect_MsSqlServer.cs
+++ b/src/SqlPersistence/Saga/SqlDialect_MsSqlServer.cs
@@ -20,10 +20,6 @@ namespace NServiceBus
                 return writer.ToCharSegment();
             }
 
-            internal override void ValidateJsonSettings(Newtonsoft.Json.JsonSerializer jsonSerializer)
-            {
-            }
-
             public override string BuildSaveCommand(string correlationProperty, string transitionalCorrelationProperty, string tableName)
             {
                 var valuesBuilder = new StringBuilder();
--- a/src/SqlPersistence/Saga/SqlDialect_MsSqlServer.cs
+++ b/src/SqlPersistence/Saga/SqlDialect_MsSqlServer.cs
@@ -20,6 +20,10 @@ namespace NServiceBus
                 return writer.ToCharSegment();
             }
 
+            internal override void ValidateJsonSettings(Newtonsoft.Json.JsonSerializer jsonSerializer)
+            {
+            }
+
             public override string BuildSaveCommand(string correlationProperty, string transitionalCorrelationProperty, string tableName)
             {
                 var valuesBuilder = new StringBuilder();
--- a/src/SqlPersistence/Saga/SqlDialect_MySql.cs
+++ b/src/SqlPersistence/Saga/SqlDialect_MySql.cs
@@ -13,6 +13,10 @@ namespace NServiceBus
                 return $"`{tablePrefix}{tableSuffix}`";
             }
 
+            internal override void ValidateJsonSettings(Newtonsoft.Json.JsonSerializer jsonSerializer)
+            {
+            }
+
             public override string BuildSaveCommand(string correlationProperty, string transitionalCorrelationProperty, string tableName)
             {
                 var valuesBuilder = new StringBuilder();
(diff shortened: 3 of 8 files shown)


SzymonPobiega:
Why do we even allow customization of the serializer? Do we know any valid use cases?

SimonCropp:
> Why do we even allow customization of the serializer

yes https://docs.particular.net/persistence/sql/saga#json-net-settings

> Do we know any valid use cases

they own the saga data structure and need to have control over how it is serialized. also exposing the settings allows a more expressive way of evolving the shaper of sagas over time

src/SqlPersistence/Saga/SqlDialect_PostgreSql.cs
```diff
@@ -3,16 +3,43 @@
...
             public override string GetSagaTableName(string tablePrefix, string tableSuffix)
             {
                 return $"\"{Schema}\".\"{tablePrefix}{tableSuffix}\"";
             }
 
+            internal override void ValidateJsonSettings(Newtonsoft.Json.JsonSerializer jsonSerializer)
+            {
+                if (!IsTypeNameHandlingCorrect(jsonSerializer))
+                {
+                    throw new SerializationException("PostgreSQL does not guarantee that properties are stored in order. As such, when using any TypeNameHandling other than 'None', then the MetadataPropertyHandling must be set to 'ReadAhead'.");
```
danielmarbach: it feels a bit off that we use here the SerializationException since we are not actually serializing anything. Why not just a standard exception?

src/SqlPersistence/Saga/SqlDialect.cs
```diff
@@ -24,5 +24,7 @@ internal virtual object BuildSagaData(CommandWrapper command, RuntimeSagaInfo sa
         {
             return sagaInfo.ToJson(sagaData);
         }
+
+        internal abstract void ValidateJsonSettings(Newtonsoft.Json.JsonSerializer jsonSerializer);
```
danielmarbach: Why not make this virtual?

danielmarbach: Is there a reason the serializer is fully qualified here?

src/SqlPersistence/Saga/SqlDialect_MsSqlServer.cs
```diff
@@ -20,6 +20,10 @@ internal override object BuildSagaData(CommandWrapper command, RuntimeSagaInfo s
                 return writer.ToCharSegment();
             }
 
+            internal override void ValidateJsonSettings(Newtonsoft.Json.JsonSerializer jsonSerializer)
```
danielmarbach: Is there a reason the serializer is fully qualified here?

src/SqlPersistence/Saga/SqlDialect_MySql.cs
```diff
@@ -13,6 +13,10 @@ public override string GetSagaTableName(string tablePrefix, string tableSuffix)
                 return $"`{tablePrefix}{tableSuffix}`";
             }
 
+            internal override void ValidateJsonSettings(Newtonsoft.Json.JsonSerializer jsonSerializer)
```
danielmarbach: Is there a reason the serializer is fully qualified here?

src/SqlPersistence/Saga/SqlDialect_Oracle.cs
```diff
@@ -22,6 +22,10 @@ public override string GetSagaTableName(string tablePrefix, string tableSuffix)
                 return tableSuffix.ToUpper();
             }
 
+            internal override void ValidateJsonSettings(Newtonsoft.Json.JsonSerializer jsonSerializer)
```
danielmarbach: Is there a reason the serializer is fully qualified here?

src/SqlPersistence/Saga/SqlDialect_PostgreSql.cs
```diff
@@ -3,16 +3,43 @@
...
         public partial class PostgreSql
         {
+            static ILog log = LogManager.GetLogger<PostgreSql>();
+
             public override string GetSagaTableName(string tablePrefix, string tableSuffix)
             {
                 return $"\"{Schema}\".\"{tablePrefix}{tableSuffix}\"";
             }
 
+            internal override void ValidateJsonSettings(Newtonsoft.Json.JsonSerializer jsonSerializer)
```
danielmarbach: Is there a reason the serializer is fully qualified here?

src/SqlPersistence/SerializationException.cs
```diff
@@ -7,5 +7,8 @@ class SerializationException : Exception
         public SerializationException(Exception innerException) : base("Serialization failed", innerException)
         {
         }
+        public SerializationException(string message) : base(message)
```
danielmarbach: This could go when the code would not throw a serialization exception for the misconfiguration

src/SqlPersistence/Saga/SqlDialect_PostgreSql.cs
weralabaj: I'm not sure if that's a good argument, but it is validation closely related to serialization. May make it easier for the user to understand the source of the problem. No hard feelings around changing to generic "Exception" or "ArgumentException" maybe?

weralabaj: Simon says "Also cant be a generic exception since we want a custom exception that is excluded from retries" so better leave as is IMO

src/SqlPersistence/Saga/SqlDialect.cs
weralabaj: clash with NServiceBus.JsonSerializer

src/SqlPersistence/Saga/SqlDialect_PostgreSql.cs
weralabaj: name clash

src/SqlPersistence/Saga/SqlDialect_PostgreSql.cs
danielmarbach: Are you saying this method is only executed during runtime and not during bootstrapping? Then it make sense

PR #226, opened 2017-11-16
Project cleanup

bording:
This PR makes changes to the project files to clean up aspects of moving to the new project system (including simplifying all of the dependencies to take advantage of transitive references), moves to using Particular.Packaging, and adds SourceLink support.

I see that the postgresql branch is also based against develop, and it looks like there will be some overlap there.

Is that branch being merged for 3.0 as well? I'd be happy to help integrate the two sets of changes.


src/ScriptBuilder/ResourceReader.cs
```diff
@@ -8,7 +8,7 @@ static class ResourceReader
 
     public static string ReadResource(BuildSqlDialect sqlDialect, string prefix)
     {
-        var text = $"NServiceBus.Persistence.Sql.{prefix}_{sqlDialect}.sql";
+        var text = $"NServiceBus.Persistence.Sql.ScriptBuilder.{prefix}_{sqlDialect}.sql";
```
DavidBoike: Did all the resources move?

src/ScriptBuilderTask/ScriptBuilderTask.csproj
```diff
@@ -1,23 +1,24 @@
...
     <TargetFrameworks>net452;netstandard1.5</TargetFrameworks>
     <AssemblyName>NServiceBus.Persistence.Sql.ScriptBuilderTask</AssemblyName>
     <RootNamespace>NServiceBus.Persistence.Sql</RootNamespace>
-    <AssemblyTitle>NServiceBus.Persistence.Sql.ScriptBuilderTask</AssemblyTitle>
-    <Product>NServiceBus.Persistence.Sql.ScriptBuilderTask</Product>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>$(SolutionDir)NServiceBus.snk</AssemblyOriginatorKeyFile>
     <DebugType>full</DebugType>
     <OutputPath>..\..\binaries\</OutputPath>
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
```
DavidBoike: I had to look this up - it silences "Missing XML comment for publicly visible type or member" for ScriptBuilderTask.

src/SqlPersistence/SqlPersistence.csproj
```diff
@@ -1,51 +1,44 @@
...
     <TargetFrameworks>net452;netstandard2.0</TargetFrameworks>
     <AssemblyName>NServiceBus.Persistence.Sql</AssemblyName>
     <RootNamespace>NServiceBus.Persistence.Sql</RootNamespace>
-    <AssemblyTitle>NServiceBus.Persistence.Sql</AssemblyTitle>
-    <Product>NServiceBus.Persistence.Sql</Product>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>$(SolutionDir)NServiceBus.snk</AssemblyOriginatorKeyFile>
-    <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <OutputPath>..\..\binaries\</OutputPath>
-    <NoWarn>$(NoWarn);CS0419</NoWarn>
```
DavidBoike: CS0419 = Ambiguous reference in cref attribute

src/Integration.sln
```diff
@@ -10,6 +10,11 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SqlServerSample", "Integrat
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OracleSample", "Integration\OracleSample\OracleSample.csproj", "{E7089FA4-6E0E-443E-B6B3-86D07B092A52}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{59006A92-CC0F-4A5C-9CE2-D43BC11882CB}"
+	ProjectSection(SolutionItems) = preProject
+		Integration\Directory.Build.targets = Integration\Directory.Build.targets
```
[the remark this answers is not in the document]
bording: That's just a solution item so you can see/edit the file in VS.

src/ScriptBuilder/ResourceReader.cs
bording: The default namespace for the project, set via the RootNamespace project property, was set to `NServiceBus.Persistence.Sql` instead of `NServiceBus.Persistence.Sql.ScriptBuilder`, which is the actual namespace being used in the project, so I changed it the property to reflect the actual value.

It appears that the default namespace is used when embedding resources, so this needed to change as well to accommodate the namespace change.

src/ScriptBuilderTask/ScriptBuilderTask.csproj
bording: I had to add that because using Particular.Packaging will cause a project to create an xml documentation file, and this project wasn't doing that before, so it started spitting out warnings.

src/SqlPersistence/SqlPersistence.csproj
bording: No warnings showed up when I removed this, so I left it out.

PR #230, opened 2017-11-18
Cleanup followup

bording:
Because of some bad gitattributes settings, the merge of the postgres support didn't correctly mark some files as conflicting and instead just merged in the changes on both sides. I've got through and fixed up everything.

I also cleaned up the two new project files to match the others.

One last thing I did was to change how the new command tests are prevented from running on netcoreapp2.0. I switched to conditional compilation because if either oracle or mysql started working on .NET Core, those tests would still need to be excluded because they use ApprovalTests.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -64,11 +64,6 @@
 *.bat text eol=crlf
 
 # Custom for Visual Studio
-*.csproj merge=union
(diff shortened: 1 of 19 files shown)


SzymonPobiega:
Thanks @bording ❤️

PR #241, opened 2017-12-08
Prepare for final beta release

bording:


bording:
There have been things I've needed to fix that appear to be more regressions. There are still test failures that appear to indicate something isn't right with the Oracle support.

SzymonPobiega:
@bording I think there is a schema-related problem with Oracle tests on TC. I'll investigate. I think we're missing a schema on that DB. Merging this as the problem is unrelated.

bording:
@SzymonPobiega In addition to any problem there might be with the Oracle server on TC, note the following tests are failing because of approval file problems:

- SagaScriptBuilderTest.CreateWithCorrelation(Oracle) 
- SagaScriptBuilderTest.CreateWithCorrelationAndTransitional(Oracle)

Which seems to indicate there might be something more going on than just a TC build agent problem.

PR #242, opened 2017-12-12
[WIP] Hotfix 3.0.2

SzymonPobiega:
Do not merge.


DavidBoike:
TC is trying to build it as 4.1.0 with a commit count of 240? It seems to be ridiculously out of whack in how it's looking at commits in the repository. Was history changed enough that we need to somehow force TC to do a clean clone?

SzymonPobiega:
I think this is a known problem: TC builds PRs as if they were for develop,
Not the actual target branch.

PR #252, opened 2018-01-15
make script execution a public api

SimonCropp:
so that people can reuse the code without the need to start a full endpoint


SzymonPobiega:
@SimonCropp Thanks. Looks good. Let me add a doco PR and then I'll merge it.

SimonCropp:
@SzymonPobiega can we get the same for the sql transport?

SimonCropp:
also can we get a new nuget

DavidBoike:
@SimonCropp Beta4 pushed to MyGet and (in theory) to NuGet, though those pushes have been ... not fast lately.

SimonCropp:
@DavidBoike 

thanks 

> though those pushes have been ... not fast lately.

yeah i have noticed on my side projects :) between 5min and 5 hours for nuget to update its index

SimonCropp:
@DavidBoike 
 just to confirm, did u also promote the msbuild package?

DavidBoike:
Whoops! Nope, forgot those other two. Just pushed them now.

SzymonPobiega:
>  can we get the same for the sql transport?

I need to talk to the SQL Server transport maintainers :trollface: Yeah, I guess it is simple enough to just do it.

SzymonPobiega:
@SimonCropp I reviewed the code and it seems like there is very little value in adding public API for it. We could only expose the code from [this line](https://github.com/Particular/NServiceBus.SqlServer/blob/release-4.0/src/NServiceBus.SqlServer/Receiving/QueueCreator.cs#L37) forward in the `QueueCreator`. The higher level code depends on the way SQL Server transport represents addresses and I'd rather not make it public.

And in that method in `QueueCreator` there is really not much code. If you want to bypass the queue creation mechanism you can always use `SqlConstants` which is a public type (sort of).

SimonCropp:
fair enough. can we make SqlConstants a supported api in a future version?

SzymonPobiega:
@SimonCropp yeah. I think that's reasonable.

PR #257, opened 2018-02-13
Fix regression in publish behavior

bording:
This fixes the regression introduced in 4.0.0-beta002.

The task expects either a valid solution directory or an empty string, so this prevents `*Undefined*` from being passed in.


bording:
@timbussmann FYI we should get this merged and tag it as RC2.

SimonCropp:
@bording BTW u should be able to test this by running, as part of the TC build, a build of a project, ie directly on the csproj and not the sln. so as to prevent a future regression

Issue #199, opened 2017-10-20
Saga left behind when completed


markgould:
Not sure if this is more appropriate here or in the NServiceBus repo. It may simply be an architecture issue on our end.

I'm debugging a strange issue where a saga completes and doesn't clean up data. I've verified it is hitting MarkAsComplete and the Saga Audit in ServiceInsight shows the Saga has completed. If I query the saga table I see a row for the saga with a Concurrency value of 9.

If I send a new message later with the same correlation id, it picks up the old saga instance rather than starting a new one. 

We have a legacy system that can fire multiple events that need to be aggregated together, and we handle these with timeouts (5 seconds) and aggregate them into one message before publishing to other systems.

We get some fairly typical "Cannot insert duplicate key row in object" in log messages initially, which I think makes sense as these messages are trying to be processed at the same time.

Should I not assume that MarkAsComplete will clean up the saga data? I notice the summary says "Marks the saga as complete. This may result in the sagas state being deleted by the persister."  

Here is the code: https://gist.github.com/markgould/edad163e60846099f4a78ebfcfddb82f

The messages get sent from a SQL stored procedure, which is firing from an insert trigger. Our legacy system can insert a number of records in a table. The stored procedure has some logic to know how to group these, and assigns it a Refund Id. The first time the stored procedure sees it and assigns a Refund Id, it sends a RefundCreated message, the next time, it is able to see a Refund Id has already been assigned and sends a RefundAmountChanged message for that Refund Id.

The saga uses a 5 second timeout to aggregate these and publishes a RefundInitiated event. We determine if this was a new refund by storing the create message on the saga.  If an adjustment happens at some point later, in theory it should be in a new saga and will only have a RefundAmountChangedMessage and publish a RefundAmountChanged event, but because it's picking up the previous saga instance, it's publishing a RefundInitiated message again.

markgould:
Should have mentioned:
NServiceBus 6.3.6
NServiceBus.Persistence.Sql 2.1.3
NServiceBus.SqlServer 3.1.2

SzymonPobiega:
@markgould  do you see concurrency exceptions on the timeout that completes the saga? What transport is the endpoint using. The reason I am asking is following.

[...]

PR #261, opened 2018-03-03
Fixes concurrency issue when trying to complete saga

markgould:
References #199 and this thread. https://discuss.particular.net/t/sql-persistence-saga-race-condition-issue/416/5

I noticed the complete logic is not checking the affected rows like the update is, so it can silently fail to delete the saga.


SzymonPobiega:
@markgould I took liberty to re-target the PR

markgould:
Thanks, appreciate the quick turnaround!

SzymonPobiega:
@markgould I've pushed the 3.0.5 to nuget. We're working on polishing the official announcement. Thanks!

PR #267, opened 2018-03-22
3.0.6 multitenant

mikeknoll:
NHibernate persistence supports different connection strings per StorageType, as demonstrated: https://docs.particular.net/samples/outbox/multi-tenant/

SQL Persistence only specify one ConnectionBuilder for all StorageTypes.

This PR adds the ability to specify a ConnectionBuilder for each StorageType, as well as a "default" ConnectionBuilder.  With seperate ConnectionBuilders, seperate connection strings can be supplied, allowing SQL Persistence to be used similarly to NHibernate's multi-tenant setup.


SzymonPobiega:
Hi @mikeknoll 

Unfortunately we can't accept this PR right now. We're still in the process of releasing NServiceBus 8 with SQL Persistence 4 and we don't work on new features right now.

The support for multi-tenant deployments with queue sharing is high on our priorities list. When the dust settles after the V8 release and we decide to prioritize this feature, we could work with you if you're still interested.

SimonCropp:
@SzymonPobiega how long till "the dust settles"?

SzymonPobiega:
@SimonCropp we don't yet know. Do you need this feature?

SimonCropp:
@SzymonPobiega nope i dont need it. but, over the past few months, I have noticed the response to a growing number of bugs and feature has been as u say "When the dust settles after the V7 release and we decide to prioritize this". just wondering when this state of limbo can be expected to end?

SzymonPobiega:
@SimonCropp the V7 dust is not an obstacle any more. I've raised an issue to manage the process of merging @mikeknoll 's changes here https://github.com/Particular/NServiceBus.Persistence.Sql/issues/276. This is likely to be included in the next minor release.

weralabaj:
@mikeknoll Sorry for the delay, but one thing that is not clear for me is the motivation for this feature. Is that to make SQL persistence similar to NHibernate persistance? 
Or do you have a deployment where that feature is needed? Why exactly?
Thanks in advance for the answer, that will help me to ensure we'll provide what you're looking for.

mikeknoll:
We have multiple applications that access a shared database, as well as per-application databases.  We use the outbox feature to access the shared database reliably.  Outbox requires the outbox data be stored in the same database as the business data.  Since they're different applications, we can't share the subscription persistence data.  So, the outbox data needs to be stored in the shared database, but the subscriptions need to be stored in the application database.
(Our current work-around is to configure endpoints that connect to the shared database to not persistence subscriptions, and handle only commands)

tmasternak:
@mikeknoll in that case would it be enough for your case if the persister supported custom connection string for subscription storage?

SzymonPobiega:
@tmasternak @weralabaj if we had this, it would also solve the problem @markgould has here https://github.com/Particular/NServiceBus.SqlServer/issues/438

weralabaj:
@mikeknoll FYI we've just released version 4.1.1 which allows specifying separate connection strings for timeouts and subscriptions https://github.com/Particular/NServiceBus.Persistence.Sql/releases/tag/4.1.1. I hope that resolves the issue we discussed on a call some time ago. Please, feel free to reach out to us directly and provide any feedback you have, either positive or negative, regarding this feature :)

PR #268, opened 2018-03-23
Use Newtonsoft.Json 11 in V3

SzymonPobiega:
I had to add reference to json in the test projects because ApprovalTests is still on json 10.


src/ScriptBuilder.Tests/ScriptBuilder.Tests.csproj
```diff
@@ -14,6 +14,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="[11.0.1, 12.0.0)" />
```
danielmarbach: Why are these additions necessary? Plus tests would not use the range but 11.0.1 directly

src/ScriptBuilderTask.Tests/ScriptBuilderTask.Tests.csproj
```diff
@@ -14,6 +14,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="[11.0.1, 12.0.0)" />
```
danielmarbach: Why are these additions necessary? Plus tests would not use the range but 11.0.1 directly

src/SqlPersistence.Tests/SqlPersistence.Tests.csproj
```diff
@@ -15,6 +15,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="[11.0.1, 12.0.0)" />
```
danielmarbach: Why are these additions necessary? Plus tests would not use the range but 11.0.1 directly

src/ScriptBuilder.Tests/CecilExtensions/CustomAttributeMock.cs
```diff
@@ -32,10 +32,13 @@ TypeReference GetTypeReference(object value)
...
     public bool HasProperties => true;
+    public bool HasConstructorArguments => false;
     public Collection<CustomAttributeNamedArgument> Fields => throw new NotImplementedException();
 
     public Collection<CustomAttributeNamedArgument> Properties
     {
         get;
     }
+
+    public Collection<CustomAttributeArgument> ConstructorArguments => new Collection<CustomAttributeArgument>();
```
danielmarbach: How is this related to Newtonsoft update?

src/ScriptBuilder.Tests/ScriptBuilder.Tests.csproj
danielmarbach: Ah saw it now in the PR description. But still the test should not use the range

src/ScriptBuilder.Tests/CecilExtensions/CustomAttributeMock.cs
SzymonPobiega: Nope but we have a wildcard dependency on mono.cecil and they released a new breaking version.

src/ScriptBuilder.Tests/ScriptBuilder.Tests.csproj
SzymonPobiega: Fixed to 11.0.1

src/SqlPersistence.Tests/SqlPersistence.Tests.csproj
SzymonPobiega: Fixed to 11.0.1

src/ScriptBuilderTask.Tests/ScriptBuilderTask.Tests.csproj
SzymonPobiega: Fixed to 11.0.1

src/ScriptBuilder.Tests/CecilExtensions/CustomAttributeMock.cs
danielmarbach: Would it make sense to lock in the version?

SzymonPobiega: I am not sure. We didn't have problems with this dependency so far.

PR #271, opened 2018-04-02
Update targets file to include scripts in publish output

bording:
This change updates the ScriptBuilder MSBuild targets to ensure that the scripts are copied to the publish output when publishing an SDK-style project from VS 2017 or from the commandline via `dotnet publish`.

### DIFF
--- a/src/ScriptBuilderTask/NServiceBus.Persistence.Sql.MsBuild.targets
+++ b/src/ScriptBuilderTask/NServiceBus.Persistence.Sql.MsBuild.targets
@@ -34,15 +34,33 @@
       <SqlPersistenceScripts Include="$(SqlPersistenceScriptDirectory)**\*.sql" />
     </ItemGroup>
 
-    <CreateItem Include="@(SqlPersistenceScripts)"
-                AdditionalMetadata="CopyToOutputDirectory=PreserveNewest;TargetPath=NServiceBus.Persistence.Sql\%(RecursiveDir)%(Filename)%(Extension)">
-      <Output TaskParameter="Include" ItemName="_SourceItemsToCopyToOutputDirectoryAlways" />
-    </CreateItem>
+    <ItemGroup>
+      <_SourceItemsToCopyToOutputDirectoryAlways Include="@(SqlPersistenceScripts)">
+        <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+        <TargetPath>NServiceBus.Persistence.Sql\%(RecursiveDir)%(Filename)%(Extension)</TargetPath>
+      </_SourceItemsToCopyToOutputDirectoryAlways>
+    </ItemGroup>
 
-    <CreateItem Include="@(SqlPersistenceScripts)"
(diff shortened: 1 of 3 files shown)


src/ScriptBuilderTask/NServiceBus.Persistence.Sql.MsBuild.targets
```diff
@@ -34,15 +34,33 @@
...
+      <SqlPersistenceScriptDirectory>$(ProjectDir)$(IntermediateOutputPath)NServiceBus.Persistence.Sql\</SqlPersistenceScriptDirectory>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <SqlPersistenceScriptsForPublish Include="$(SqlPersistenceScriptDirectory)**\*.sql" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <_SourceItemsToCopyToPublishDirectoryAlways Include="@(SqlPersistenceScriptsForPublish)">
+        <CopyToPublishDirectory>Alway</CopyToPublishDirectory>
```
DavidBoike: Alway ???

src/ScriptBuilderTask/NServiceBus.Persistence.Sql.MsBuild.targets
```diff
@@ -34,15 +34,33 @@
...
+    <PropertyGroup>
+      <SqlPersistenceScriptDirectory>$(ProjectDir)$(IntermediateOutputPath)NServiceBus.Persistence.Sql\</SqlPersistenceScriptDirectory>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <SqlPersistenceScriptsForPublish Include="$(SqlPersistenceScriptDirectory)**\*.sql" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <_SourceItemsToCopyToPublishDirectoryAlways Include="@(SqlPersistenceScriptsForPublish)">
```
DavidBoike: Is this underscore notation a dirty hack or actually a thing?

src/ScriptBuilderTask/NServiceBus.Persistence.Sql.MsBuild.targets
bording: Ah, I thought I fixed that, must have done it to the wrong copy.

bording:
>@bording You may need to walk me through this one, but currently getting compile errors:

@DavidBoike Ah yes, wildcards. The gift that keeps on giving. 😢

src/ScriptBuilderTask/NServiceBus.Persistence.Sql.MsBuild.targets
bording: It's the way MSBuild targets indicate that the item is "internal" since everything is actually public in MSBuild.

That does mean there's a chance that could change in a future release of the targets files that implement the `GetCopyToOutputDirectoryItems` and `GetCopyToPublishDirectoryItems` targets, but I don't see any other way to hook into them.

PR #275, opened 2018-05-02
use bool comparison

SimonCropp:


SzymonPobiega:
@bording can you review this?

bording:
I would recommend not making this change. It's standard practice in MSBuild to put comparisons in quotes like I've done here. From what I understand the reason is there are some cases where they aren't required, but there are cases where they are, so its better to just always use them.

SimonCropp:
@bording i got the MS guidance changed https://github.com/dotnet/project-system/pull/3510

PR #279, opened 2018-05-11
Add test from SQL Server transport

bording:
This adds the test removed in https://github.com/Particular/NServiceBus.SqlServer/pull/440.

Looking at what this test is doing, is this scenario already covered by the other tests in the integration folder? If so, then I guess we don't need to add this one.


bording:
FYI I've already fixed the inspection failures in #278, so it will be okay to merge this with the failures and they will be fixed over there.

bording:
Seems like there's something going on in the test that doesn't like the other tests in the project, but I'm not sure what needs to change.

bording:
Closing this PR because @SzymonPobiega is going to address this scenario in a different way in a PR targeting `develop` instead.

Issue #276, opened 2018-05-07
Allow SQL persistence to use different connection strings for subscriptions and timeouts


SzymonPobiega:
Update: [Here](https://gist.github.com/SzymonPobiega/49e3d1b340755dee8edbfb2dc93725c1) the code to smoke test this feature as it is super hard to test it automatically (requires separate DBs etc.)

Allow SQL persistence to use different data stores for different persistence concerns e.g. different database for subscriptions and for sagas.

New API proposal:

```
//Current API
var persistence = config.UsePersistence<SqlPersistence>();
persistence.ConnectionBuilder(() => new SqlConnection())

//New API
var persistence = config.UsePersistence<SqlPersistence>();
var subscriptions = persistence.SubscriptionSettings();
subscriptions.ConnectionBuilder(() => new SqlConnection())

var timeouts = persistence.TimeoutSettings();
timeouts.ConnectionBuilder(() => new SqlConnection())

Customer-provided implementation is here: https://github.com/Particular/NServiceBus.Persistence.Sql/pull/267

SzymonPobiega:
@Particular/sqlpersistence-maintainers I propose this to be part of the next maintenance release

SzymonPobiega:
@DavidBoike @weralabaj @tmasternak can you review the API I proposed? I hope it covers the saga/outbox relation well enough. The rules here are a bit tricky:
 * When the outbox is enabled, the outbox connection is used for sagas so sagas can't specify their connection explicitly. If you want to customize where the sagas are stored, you customize the *outbox* connection factory
 * When the outbox is disabled and you want to customize where the sagas are stored, you customize the *sagas* connection factory
 * When the outbox is disable you can also use a connection factory that takes the message context so that you can store your sagas based on the message headers

I could not find a way to express these rules in a type-safe manner. I think throwing an exception if someone provides a saga connection when the outbox is enabled is a good enough solution.

weralabaj:
Also 👍 for illegal configuration, although I have a bit of a challenge to understand what scenarios we're covering and why.

I get the argument about storing all subscriptions in a single database, but I'm not sure why sagas would be stored in various databases?

Is it still a valid scenario since we moved away from a multi-database setup in SQL transport?

tmasternak:
> Is it still a valid scenario since we moved away from a multi-database setup in SQL transport?

I think what we want to tackle is:
 * single db for messaging,
 * per-tenant db for saga data.

[...]

PR #284, opened 2018-05-21
Override connection for subscriptions

SzymonPobiega:
Related to https://github.com/Particular/NServiceBus.Persistence.Sql/issues/276

[Here](https://gist.github.com/SzymonPobiega/49e3d1b340755dee8edbfb2dc93725c1) the code to smoke test this feature as it is super hard to test it automatically (requires separate DBs etc.)


src/SqlPersistence/Config/SqlPersistenceConfig_Connection.cs
```diff
@@ -19,14 +19,24 @@ public static void ConnectionBuilder(this PersistenceExtensions<SqlPersistence>
...
-            if (settings.TryGet("SqlPersistence.ConnectionBuilder", out Func<DbConnection> value))
+            if (settings.TryGet($"SqlPersistence.ConnectionBuilder.{storageType.Name}", out Func<DbConnection> value))
+            {
+                return value;
+            }
+            if (settings.TryGet("SqlPersistence.ConnectionBuilder", out value))
             {
                 return value;
             }
             throw new Exception("ConnectionBuilder must be defined.");
```
DavidBoike: I know this wasn't changed but maybe we could provide a better exception here, and also include the storage type it's trying to find.

src/SqlPersistence/ScriptRunner.cs
```diff
@@ -29,30 +40,37 @@ public static class ScriptRunner
...
+            using (var connection = connectionBuilder(typeof(T)))
+            {
+                await connection.OpenAsync().ConfigureAwait(false);
+                using (var transaction = connection.BeginTransaction())
                 {
-                    await InstallTimeouts(scriptDirectory, connection, transaction, tablePrefix, sqlDialect).ConfigureAwait(false);
+                    await installAction(scriptDirectory, connection, transaction, tablePrefix, sqlDialect).ConfigureAwait(false);
                 }
-
-                transaction.Commit();
```
DavidBoike: Should we worry that the installers won't run transactionally? Or, vs. the cost of trying to group things by connection builder and somehow execute common ones together in a transaction still?

DavidBoike:
@SzymonPobiega Busted by APIApprover :wink:

src/SqlPersistence/Config/SqlPersistenceConfig_Connection.cs
weralabaj: > also include the storage type it's trying to find.

👍

src/SqlPersistence/InstallerFeature.cs
```diff
@@ -16,7 +16,11 @@ protected override void Setup(FeatureConfigurationContext context)
             return;
         }
 
-        settings.ConnectionBuilder = context.Settings.GetConnectionBuilder();
+        settings.ConnectionBuilder = storageType =>
```
weralabaj: Could be `settings.ConnectionBuilder = storageType => context.Settings.GetConnectionBuilder(storageType)();` or you keep it that way for readability?

If that was the explicit intention then ignore comment.

src/SqlPersistence/ScriptRunner.cs
weralabaj: I think this is a bug? I don't see where those transactions are commited.

SzymonPobiega: I don't think have a single transaction is necessary here as these installers are not destructive.

src/SqlPersistence/ScriptRunner.cs
```diff
@@ -29,30 +40,37 @@ public static class ScriptRunner
...
 
-            using (var connection = await connectionBuilder.OpenConnection().ConfigureAwait(false))
-            using (var transaction = connection.BeginTransaction())
+            if (shouldInstallOutbox)
             {
-                if (shouldInstallOutbox)
-                {
-                    await InstallOutbox(scriptDirectory, connection, transaction, tablePrefix, sqlDialect).ConfigureAwait(false);
-                }
+                await Execute<StorageType.Outbox>(InstallOutbox, scriptDirectory, tablePrefix, sqlDialect, connectionBuilder).ConfigureAwait(false);
```
weralabaj: I know this is used to avoid repetition, but I'd prefer to have that "execute" method inlined. It's really hard to follow this code :/

Or, the minimum would be to give this method a more descriptive name like `ExecuteScriptTransactionally` or something like that.

src/SqlPersistence/Config/SqlPersistenceConfig_Connection.cs
```diff
@@ -19,13 +19,23 @@ public static void ConnectionBuilder(this PersistenceExtensions<SqlPersistence>
...
+            if (settings.TryGet($"SqlPersistence.ConnectionBuilder.{storageType.Name}", out Func<DbConnection> value))
             {
                 return value;
             }
-            throw new Exception("ConnectionBuilder must be defined.");
+            if (settings.TryGet("SqlPersistence.ConnectionBuilder", out value))
+            {
+                return value;
+            }
+            throw new Exception("The connection to the database must be specified using the ConnectionBuilder method.");
```
weralabaj: Suggestion:
` throw new Exception($"Couldn't find connection string for {storageType}. The connection to the database must be specified using the ConnectionBuilder method.");`

src/SqlPersistence/InstallerFeature.cs
weralabaj: ping

src/SqlPersistence/ScriptRunner.cs
weralabaj: ping

SzymonPobiega: I am going with `ExecuteInSeparateConnection`.

SzymonPobiega:
@weralabaj I applied your suggestions

PR #280, opened 2018-05-15
Move integration tests to separate project

SzymonPobiega:

PR #285, opened 2018-05-21
Misc cleanup

bording:
Here is a follow-up PR to #280 that cleans up a few unneeded things that were introduced there. It also removes the references to Microsoft.CSharp, which aren't needed.


weralabaj:
@bording That should go with the .NET Core release, correct?

bording:
@weralabaj I don't understand your question.

PR #286, opened 2018-05-22
Use pessimistic locking when loading sagas to prevent lock contention on congested sagas

SzymonPobiega:
* Adds `for update` or equivalent clause to all three saga loading SQLs: by ID, by property and by finder
 * Slightly changes the API of a dialect to allow adding the `for update` at the end of the statement

### DIFF
--- a/src/SqlPersistence.Tests/APIApprovals.Approve.approved.txt
+++ b/src/SqlPersistence.Tests/APIApprovals.Approve.approved.txt
@@ -133,7 +133,7 @@ namespace NServiceBus
         [System.ObsoleteAttribute("Not for public use")]
         public abstract string BuildSaveCommand(string correlationProperty, string transitionalCorrelationProperty, string tableName);
         [System.ObsoleteAttribute("Not for public use")]
-        public abstract string BuildSelectFromCommand(string tableName);
+        public abstract System.Func<string, string> BuildSelectFromCommand(string tableName);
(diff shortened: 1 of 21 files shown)


DavidBoike:
Is the `where 1 = 1` stuff serve a SQL API purpose or is it just to make the select commands easier to generate? If the latter, I'm a little worried a DBA might look at that and scoff.

weralabaj:
@DavidBoike I see this is only used as a "test condition" in tests, so this isn't part of the package code.

SzymonPobiega:
@weralabaj @DavidBoike yeah, the `1=1` is only used in tests to make sure that the value is inserted in the right place in the generated SQL

SzymonPobiega:
> Why did you add "where" parameter and changed the methods signatures

@weralabaj I had to do it because the finder query was being assembled outside of the dialect by appending the `where blabla` clause. Because the `for update` clause needs to be the last thing in the SQL and it needs to be handled in the dialect -> dialect needs to also append the `where`

SzymonPobiega:
> In doco we need to explicitly state what level of locking we use and what impact it has for isolation level.

True. I don't think we have any doco regarding the isolation levels supported :-(

weralabaj:
> True. I don't think we have any doco regarding the isolation levels supported :-(

We don't, just making a note that we should cover that gap ;)

PR #287, opened 2018-05-22
WIP Test MySQL with MySQL.Data 8.x on .NET Core

DavidBoike:


DavidBoike:
It works! Going to close this PR and open a new PR against master with the squashed commits. Then that branch can be merged into master and then into develop, to keep TeamCity happy on all support branches, as getting the MySQL tests to run for .NET Core involved changes to TeamCity configuration as well.

DavidBoike:
Strike that, the successful test run had 343 successful tests - before adding MySQL it was 313.

.NET Framework has 882 with MySQL+Oracle. So I would have expected in the 500-600 range. Guessing MySQL acceptance tests aren't being found/executed. Will keep digging.

DavidBoike:
Nevermind, this is a reporting problem, not a testing problem. Just as much a problem already with Postgres as it is with MySQL. The code is good.

PR #288, opened 2018-05-22
Test MySQL with MySQL.Data 8.x on .NET Core

DavidBoike:
Although there's a reporting problem on TeamCity, it looks like MySQL.Data 8.x on .NET Core works just fine. This PR adds it to master. While we could release a 4.0.1 instead of a 4.0.0, I'm not sure we even have to.

After this is merged, the same branch (or master?) can be merged into develop and then into any feature branches so that both 4.0.x and 4.1.x and the current maintenance release will have the new tests and cooperate with the changes that had to be made on TeamCity.

### DIFF
--- a/src/MySqlAcceptanceTests/MySqlAcceptanceTests.csproj
+++ b/src/MySqlAcceptanceTests/MySqlAcceptanceTests.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net452</TargetFramework>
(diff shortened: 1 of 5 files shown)


bording:
@DavidBoike I pushed up a tweak. All multi-targeted test projects need to have `<CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>` in order to guarantee that all the required assemblies are in the bin folder. Without that, the .NET Core tests on Windows are only passing because the build agents happen to have already restored the required packages before.

weralabaj:
@DavidBoike Not sure what reporting problem you have in mind?

Issue #272, opened 2018-04-12
Install actions fail silently when running via NSB host


SzymonPobiega:
When an error occurs while running the installer via NSB host the process completes without showing the error, leaving the environment partially configured.

Based on support case https://c.eu1.visual.force.com/apex/Case?num=00035886

SzymonPobiega:
@tmasternak how do we go about triaging this?

SzymonPobiega:
The NHibernate schema update feature logs error when an exception is thrown during schema update. The solution is to add similar code to SQL persistence because it can't depend on the host to log the error.

SzymonPobiega:
Closed via #291

PR #291, opened 2018-05-23
Log installer failures as error

SzymonPobiega:
Fixes https://github.com/Particular/NServiceBus.Persistence.Sql/issues/272

### DIFF
--- a/src/SqlPersistence/NeedToInstallSomething.cs
+++ b/src/SqlPersistence/NeedToInstallSomething.cs
@@ -40,8 +40,7 @@ class SqlPersistenceInstaller : INeedToInstallSomething
         }
         catch (Exception e)
         {
-            //The NServiceBus host has an empty Try Catch block and does not log the exception so we need to log it here.
-            log.Error("Could not complete the schema update", e);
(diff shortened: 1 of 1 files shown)


src/SqlPersistence/NeedToInstallSomething.cs
```diff
@@ -1,35 +1,48 @@
...
+                    shouldInstallOutbox: settings.IsFeatureActive(typeof(SqlOutboxFeature)),
+                    shouldInstallSagas: settings.IsFeatureActive(typeof(SqlSagaFeature)),
+                    shouldInstallSubscriptions: settings.IsFeatureActive(typeof(SqlSubscriptionFeature)),
+                    shouldInstallTimeouts: settings.IsFeatureActive(typeof(SqlTimeoutFeature)))
+                .ConfigureAwait(false);
+        }
+        catch (Exception e)
+        {
+            //The NServiceBus host has an empty Try Catch block and does not log the exception so we need to log it here.
+            log.Error("Could not complete the schema update", e);
```
weralabaj: Since we could have various errors here I'm not clear why we have such specific error message here?

PR #295, opened 2018-05-25
fix outbox diagnostics

weralabaj:


bording:
As mentioned elsewhere, this should be merged to master as `hotfix-4.1.1` instead.

PR #298, opened 2018-06-05
Allow use of regular Saga<T> base class

DavidBoike:
Connects to https://github.com/Particular/PlatformDevelopment/issues/2020

Enables use of normal `Saga<T>` API, throwing an error at build time if the user does anything far outside the ordinary and we can't determine correlation id from the ConfigureHowToFindSaga method, in which case, the user can either fix it, decorate with a (restored) [SqlSaga] attribute, or use `SqlSaga<T>`.

`SqlSaga<T>` is still supported and likely will continue to be so for a long time.

### DIFF
--- a/src/AcceptanceTestHelper/RuntimeSagaDefinitionReader.cs
+++ b/src/AcceptanceTestHelper/RuntimeSagaDefinitionReader.cs
@@ -5,6 +5,7 @@ using System.Reflection;
 using System.Runtime.Serialization;
 using Mono.Cecil;
 using NServiceBus;
+using NServiceBus.Persistence.Sql;
 using NServiceBus.Persistence.Sql.ScriptBuilder;
 
 public static class RuntimeSagaDefinitionReader
@@ -40,6 +41,11 @@ public static class RuntimeSagaDefinitionReader
 
     public static SagaDefinition GetSagaDefinition(Type sagaType, BuildSqlDialect sqlDialect)
     {
+        if (SagaTypeHasIntermediateBaseClass(sagaType))
+        {
+            throw new Exception("Saga implementations must inherit from either Saga<T> or SqlSaga<T> directly. Deep class hierarchies are not supported.");
+        }
+
         var saga = (Saga)FormatterServices.GetUninitializedObject(sagaType);
         var mapper = new ConfigureHowToFindSagaWithMessage();
         methodInfo.Invoke(saga, new object[]
@@ -53,10 +59,8 @@ public static class RuntimeSagaDefinitionReader
                 name: mapper.CorrelationProperty,
                 type: CorrelationPropertyTypeReader.GetCorrelationPropertyType(mapper.CorrelationType));
         }
-        var transitionalCorrelationPropertyName = (string)sagaType
-            .GetProperty("TransitionalCorrelationPropertyName", AnyInstanceMember)
-            .GetValue(saga);
-
+        
+        var transitionalCorrelationPropertyName = GetSagaMetadataProperty(sagaType, saga, "TransitionalCorrelationPropertyName", att => att.TransitionalCorrelationProperty);
 
         CorrelationProperty transitional = null;
         if (transitionalCorrelationPropertyName != null)
@@ -65,8 +69,8 @@ public static class RuntimeSagaDefinitionReader
             var transitionalProperty = sagaDataType.GetProperty(transitionalCorrelationPropertyName, AnyInstanceMember);
             transitional = new CorrelationProperty(transitionalCorrelationPropertyName, CorrelationPropertyTypeReader.GetCorrelationPropertyType(transitionalProperty.PropertyType));
         }
-
-        var tableSuffixOverride = (string)sagaType.GetProperty("TableSuffix", AnyInstanceMember).GetValue(saga);
+        
+        var tableSuffixOverride = GetSagaMetadataProperty(sagaType, saga, "TableSuffix", att => att.TableSuffix);
         var tableSuffix = tableSuffixOverride ?? sagaType.Name;
 
         if (sqlDialect == BuildSqlDialect.Oracle)
@@ -79,5 +83,37 @@ public static class RuntimeSagaDefinitionReader
             name: sagaType.FullName,
             correlationProperty: correlationProperty,
             transitionalCorrelationProperty: transitional);
+    }
+
+    static bool SagaTypeHasIntermediateBaseClass(Type sagaType)
+    {
+        var baseType = sagaType.BaseType;
+        if (!baseType.IsGenericType)
+        {
+            // Saga<T> and SqlSaga<T> are both generic types
+            return true;
+        }
+
+        var genericBase = baseType.GetGenericTypeDefinition();
+        return genericBase != typeof(Saga<>) && genericBase != typeof(SqlSaga<>);
+    }
+
+    static string GetSagaMetadataProperty(Type sagaType, Saga instance, string sqlSagaPropertyName, Func<SqlSagaAttribute, string> getSqlSagaAttributeValue)
+    {
+        if (sagaType.IsSubclassOfRawGeneric(typeof(SqlSaga<>)))
+        {
+            // ReSharper disable once PossibleNullReferenceException
+            return (string)sagaType
+                .GetProperty(sqlSagaPropertyName, AnyInstanceMember)
+                .GetValue(instance);
+        }
+
+        if (sagaType.IsSubclassOfRawGeneric(typeof(Saga<>)))
+        {
+            var attr = sagaType.GetCustomAttribute<SqlSagaAttribute>();
+            return (attr != null) ? getSqlSagaAttributeValue(attr) : null;
+        }
+
+        throw new Exception($"Type '{sagaType.FullName}' is not a Saga<T>.");
     }
 }
--- a/src/AcceptanceTestsHolder/App_Packages/NSB.AcceptanceTests/Outbox/When_clearing_saga_timeouts.cs
+++ b/src/AcceptanceTestsHolder/App_Packages/NSB.AcceptanceTests/Outbox/When_clearing_saga_timeouts.cs
@@ -7,7 +7,6 @@
     using Extensibility;
     using Features;
     using NServiceBus;
-    using NServiceBus.Persistence.Sql;
     using NServiceBus.Outbox;
     using NServiceBus.Persistence;
     using NUnit.Framework;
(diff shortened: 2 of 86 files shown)


src/SqlPersistence/Saga/SqlSagaAttribute.cs
```diff
@@ -0,0 +1,46 @@
...
+        public string TransitionalCorrelationProperty { get; }
+
+        /// <summary>
+        /// The name of the table to use when storing the current <see cref="Saga{TSagaData}"/>.
+        /// Will be appended to the value specified in <see cref="SqlPersistenceConfig.TablePrefix"/>.
+        /// </summary>
+        public string TableSuffix { get; }
+
+        /// <summary>
+        ///
```
SzymonPobiega: is the missing ctor description intentional?

src/ScriptBuilder/Saga/SagaDefinitionReader.cs
```diff
@@ -38,6 +45,192 @@ public static bool TryGetSqlSagaDefinition(TypeDefinition type, out SagaDefiniti
...
+        (
+            correlationProperty: BuildConstraintProperty(sagaDataType, correlationId),
+            transitionalCorrelationProperty: BuildConstraintProperty(sagaDataType, transitionalId),
+            tableSuffix: tableSuffix,
+            name: type.FullName
+        );
+        return true;
+    }
+
+    static string InferCorrelationId(TypeDefinition type, TypeDefinition sagaDataType)
```
SzymonPobiega: I read this method a few times and I don't feel comfortable reviewing this unless there is a test for each if statement here. Lots or arcane magic.

@bording has this brilliant idea that we should use a console tool for generating the scripts instead of a msbuild task. This way we could load the actual `dll` into the process and execute the `ConfigureHowToFindSaga` method. 

How about in this iteration we just bring back the support of `Saga` but make the attribute mandatory? The next step would be to drop the msbuild & cecil magic in favour of a simple console tool.

src/AcceptanceTestHelper/RuntimeSagaDefinitionReader.cs
```diff
@@ -80,4 +79,23 @@ public static SagaDefinition GetSagaDefinition(Type sagaType, BuildSqlDialect sq
             correlationProperty: correlationProperty,
             transitionalCorrelationProperty: transitional);
     }
+
+    static string GetSagaMetadataProperty(Type sagaType, Saga instance, string sqlSagaPropertyName, Func<SqlSagaAttribute, string> getSqlSagaAttributeValue)
+    {
+        if (sagaType.IsSubclassOfRawGeneric(typeof(SqlSaga<>)))
```
weralabaj: so is the idea that we'll run the same set of tests twice, once for core saga and once for sql saga?

asking on this method, but that's general question for all tests in general

src/AcceptanceTestsHolder/App_Packages/NSB.AcceptanceTests/Outbox/When_clearing_saga_timeouts.cs
```diff
@@ -61,18 +60,17 @@ public Task Handle(SignalDone message, IMessageHandlerContext context)
                 }
             }
 
-            public class PlaceOrderSaga : SqlSaga<PlaceOrderSaga.PlaceOrderSagaData>, IAmStartedByMessages<PlaceOrder>
+            public class PlaceOrderSaga : Saga<PlaceOrderSaga.PlaceOrderSagaData>, IAmStartedByMessages<PlaceOrder>
```
weralabaj: as above, don't we need to ensure it works the same way for both?

so keep the old one as is plus add a new test for core saga?

src/ScriptBuilder/Saga/SagaDefinitionReader.cs
```diff
@@ -38,6 +45,192 @@ public static bool TryGetSqlSagaDefinition(TypeDefinition type, out SagaDefiniti
         return true;
     }
 
+    static bool TryInferSagaDefinition(TypeDefinition type, out SagaDefinition definition)
+    {
+        var baseType = type.BaseType as GenericInstanceType;
+        definition = null;
+
+        // Class must directly inherit from NServiceBus.Saga<T> so that no tricks can be pulled from an intermediate class
```
weralabaj: should that be written down explicitly in the doco?

src/ScriptBuilder/Saga/SagaDefinitionReader.cs
```diff
@@ -38,6 +45,192 @@ public static bool TryGetSqlSagaDefinition(TypeDefinition type, out SagaDefiniti
...
+        if (baseType == null || !baseType.FullName.StartsWith("NServiceBus.Saga") || baseType.GenericArguments.Count != 1)
```
weralabaj: > !baseType.FullName.StartsWith("NServiceBus.Saga")

given namespaces reshuffling etc. should we add an explicit test that this assumption is correct for all (future) versions of core?

weralabaj: > baseType.GenericArguments.Count != 1

is that check necessary if we're already checking the name?

src/ScriptBuilder/Saga/SagaDefinitionReader.cs
```diff
@@ -1,13 +1,20 @@
...
 
 static class SagaDefinitionReader
 {
-    public static bool TryGetSqlSagaDefinition(TypeDefinition type, out SagaDefinition definition)
+    public static bool TryGetSagaDefinition(TypeDefinition type, out SagaDefinition definition)
     {
-        ValidateIsNotDirectSaga(type);
+        return TryGetSqlSagaDefinition(type, out definition)
+               || TryInferSagaDefinition(type, out definition);
+    }
```
weralabaj: > return TryGetSqlSagaDefinition(type, out definition) || TryInferSagaDefinition(type, out definition);

I think in a few months it'll make us pause for some time, can we come up with better names? maybe something like "TryGetSqlSagaDefinition ... || TryGetCoreSagaDefinition..."

src/SqlPersistence/Saga/SqlSagaAttribute.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
+
+namespace NServiceBus.Persistence.Sql
+{
+    /// <summary>
+    /// Exposes extra configuration options for storing <see cref="Saga{TSagaData}"/>.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class, Inherited = false)]
+    public sealed class SqlSagaAttribute : Attribute
```
weralabaj: Maybe I misunderstood but if we're using core saga then we shouldn't have access to those "sqlsaga-specific extras", correct?

I thought this attribute will be the (potential) next step after we implement changes in core saga to converge mappings etc.

src/ScriptBuilder/Saga/SagaDefinitionReader.cs
weralabaj: I second that opinion, this method is the direct reason why I wanted to get more eyes on this PR.

src/AcceptanceTestsHolder/App_Packages/NSB.AcceptanceTests/Outbox/When_clearing_saga_timeouts.cs
DavidBoike: No, we don't need to run any given integration style test twice. Sagas in SqlP boil down to their basic details: name, correlation id, transitional correlation id, table suffix. Basically, do saga metadata. As long as we have tests that show both styles of saga boil down to the same representation of how SqlP understands sagas (and we do) we don't need to test all of NServiceBus's possible behaviors both ways.

src/ScriptBuilder/Saga/SagaDefinitionReader.cs
DavidBoike: I will add that to the bullet list of edge cases in doco.

src/ScriptBuilder/Saga/SagaDefinitionReader.cs
DavidBoike: The SqlP package is only scoped to work with CoreV7 (or for the backport, CoreV6) and that API can't go away, so I don't know what else we should/could test?

The generic arguments count is necessary because there's `NServiceBus.Saga<T>` which we want, and then there's `NServiceBus.Saga` (no generics) which is the base class of `Saga<T>`, which we don't want.

src/ScriptBuilder/Saga/SagaDefinitionReader.cs
DavidBoike: I really really really do not want to **explicitly not solve** the problem here. This is also the first time I've been able to get *any eyes at all* actually looking at this code in the 15 months since I wrote it, so it would be really unfair to once again reject it out of hand without any opportunity to refactor.

src/SqlPersistence/Saga/SqlSagaAttribute.cs
DavidBoike: I don't see any reason to not support transitional correlation id on a core saga when it's this easy to implement. SqlSaga is a way to make correlation id mapping more DRY. I don't see a reason to overload it.

src/SqlPersistence/Saga/SqlSagaAttribute.cs
DavidBoike: I straight up copied what used to exist in the before time. :wink: I can definitely add the ctor description.

weralabaj:
@DavidBoike I think call sounds great, if it'd be easier we could also start with some pseudocode or another form of a high-level description. I think that would be helpful to ensure we covered all logical branches.

src/ScriptBuilder.Tests/Saga/CoreSagaMetadata/CoreSagaMetadataTests.cs
```diff
@@ -0,0 +1,164 @@
...
+        var dataType = module.GetTypeDefinition<DualMappingSaga>();
+        SagaDefinitionReader.TryGetSagaDefinition(dataType, out var definition);
+        ObjectApprover.VerifyWithJson(definition);
+    }
+
+    [Test]
+    public void DontMapWithIntermediateBase()
+    {
+        var dataType = module.GetTypeDefinition<HasBaseSagaClass>();
+        SagaDefinitionReader.TryGetSagaDefinition(dataType, out var definition);
```
tmasternak: Why don't this throw like all other `Dont...` type of tests?

src/SqlPersistence/Saga/SqlSagaAttribute.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        /// <summary>
+        /// Identifies the Saga correlation property if it cannot be determined by inspecting the ConfigureHowToFindSaga method.
```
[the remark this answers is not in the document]
DavidBoike: @tmasternak Updated the message to be more specific that it overrides - a "means to identify" when you can't otherwise, and that IL inspection then doesn't even happen.

DavidBoike:
Ready to review again. Set a meeting for Thursday. Additional PR to remove AcceptanceTestsHolder is now rebased in #302, which replaces #299.

src/VBTestCode/VBTestCode.vbproj
```diff
@@ -4,7 +4,7 @@
     <TargetFrameworks>net452;netcoreapp2.0</TargetFrameworks>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>$(SolutionDir)Test.snk</AssemblyOriginatorKeyFile>
-    <Optimize>false</Optimize>
+    <Optimize>true</Optimize>
```
bording: You should just drop this instead.

DavidBoike: Is true the default value?

bording: The default is configuration-specific. For Debug it's false, and for Release it's true. So, you should drop it! 😄

DavidBoike:
Rebased before merging for easier backporting but verified that the diffs were identical before pushing.

PR #301, opened 2018-06-13
Make all approval tests run on net452 and netcoreapp2.0

DavidBoike:
Removes tons of conditional compiles and gets all the approval tests running on both platforms.


DavidBoike:
@weralabaj it's because ObjectApproval 2.2.0 currently only targets .NET 4.0. We currently have no common dependency between all those tests where you could stick one copy of TestApprover (yes they're duplicates) and I didn't think it was wise to add such a dependency since it's likely in the future ObjectApproval will be updated and we could use that. I'd be happy to submit a PR but don't want to hold up the rest of the work on that.

weralabaj:
OK makes sense 👍 As for the duplication, can we extract this class to the TestHelper project?

DavidBoike:
TestHelper is only referenced by SqlPersistence.Tests and AcceptanceTestHelper (so transitively by the rest of the acceptance test projects). 

ScriptBuilderTask.Tests and ScriptBuilder.Tests (the two other places TestApprover is needed) don't reference it. TestHelper is focused on referencing the dialect client libs and providing connection builders, which really aren't needed in the two ScriptBuilder test projects. Perhaps TestHelper is poorly named?

PR #303, opened 2018-06-19
Still able to use ObjectApproval NuGet package with NoWarn="NU1701"

DavidBoike:
[Earlier PR](https://github.com/Particular/NServiceBus.Persistence.Sql/pull/301) to make approval tests cross-platform temporarily copied code from @SimonCropp's [ObjectApproval](https://www.nuget.org/packages/ObjectApproval) package as a crutch because the package doesn't (on the surface) support .NET Core, because its dependencies (ApprovalTests and ApprovalUtilities) also do not, so Simon's hands are pretty much tied there.

This PR returns to using ObjectApproval with the same `NoWarn="NU1701"` we are using for ApprovalTests and ApprovalUtilities so that it will work only because none of the packages treads on any APIs that fall outside netcoreapp2.0.

@SimonCropp I'd love to submit a PR for ApprovalTests but not sure there's really anything that can be done at the moment while the dependencies are still netfx-only?


SimonCropp:
> I'd love to submit a PR for ApprovalTests but not sure there's really anything that can be done at the moment while the dependencies are still netfx-only?

@DavidBoike while i have push access to approval tests, i am not a core maintainer. I only have push access since i have done the occasional paring session with @isidore (Llewellyn Falco).  so if you wanted to help out you need to talk to Llewellyn

PR #308, opened 2018-08-10
Fix dependencies

kbaley:
Pegged wild cards to specific versions to fix broken build


kbaley:
@Particular/sqlpersistence-maintainers This _should_ work despite the build failure in TeamCity. It's due to a GitVersion bug (which I'm told has been fixed but not released) that is picking up the version number from `develop` rather than the support branch.

PR #311, opened 2018-08-24
Update test projects to NServiceBus 7.1 and .NET Core 2.1

timbussmann:


timbussmann:
I'm not sure whether that's related, but the Integration solution seems to have some issues to build but I'm not familiar with the solution to find a quick fix. I hope that @Particular/sqlpersistence-maintainers can have a closer look at this.

PR #319, opened 2018-10-24
remove NServiceBus.Persistence.Sql.MsBuild from scriptlocation.cs

SimonCropp:


SimonCropp:
@DavidBoike thanks. i did have an afterthought. perhaps u want to mention for them to check for `<SqlPersistenceGenerateScripts>false</SqlPersistenceGenerateScripts>` in that error message?

Issue #320, opened 2018-10-31
Multi tenancy support


markgould:
This was partially addressed by #284 but does not address the scenario where one endpoint instance may be serving multiple tenants that have different databases. I.e for tenant1 the Saga and business data might be located in tenant1_db and for tenant2 it might be in tenant2_db.

Something like this would be preferable: https://docs.particular.net/samples/outbox/multi-tenant/

Right now the connectionBulider is just a `Func<DbConnection>`. If it was a `Func<ContextBag, DbConnection>` that would allow me to access the tenantId from the `ContextBag` and open the appropriate database connection. Right now that does not appear possible.

I looked at implementing it, and it is straightforward for the most part. However, this won't play nicely with the OutboxCleaner, as this runs outside of the message context, so there would be no tenant identifier at this time.

ramonsmits:
@markgould Do you mean that the approach of the multi-tenant outbox sample does not work for SQL Persistence?

Regarding the outbox cleaner, that is not multi-tenant aware and would require to know all tenant ID's. Probably a much better approach is to disable outbox cleanup in the endpoint instance and have a scheduled task run in the database server instance that does a similar query.

markgould:
@ramonsmits  - yes that example will not work for SQL Persistence.

The work that was done so far has been very useful, as you can now specify a different a different connection string for subscriptions/timeouts. We use this to keep our subscriptions in a different database from the business data. This still requires an endpoint per tenant, when tenants have different databases.

Now what we are doing is looking at moving some of our on premise customers to the cloud, we would prefer not to have to run endpoint instances per tenant. With this we are also using the Azure Service Bus transport, so subscriptions are no longer a concern.

Right now you cannot customize the connection string based on message context information. The connection builder has no parameters passed to it.

Turning off the outbox cleaner is a good idea. I am going to submit a PR soon showing one way this could work.

markgould:
@ramonsmits I was wrong, the NHibernate sample does actually work. It just feels a little odd using this AsyncLocal variable. In fact @danielmarbach has an article saying to avoid this! 

https://www.planetgeek.ch/2016/04/26/avoid-threadstatic-threadlocal-and-asynclocal-float-the-state-instead/

I will use this for now, but I would like to suggest someone look at the PR I submitted as an example of how the context information can flow through directly to the connection builder.

ramonsmits:
[...]

PR #321, opened 2018-11-01
Feature multi tenancy

markgould:
This is an example showing how multi tenancy might be implemented (See issue #320). 

Another option I looked into was replacing the ISynchronizedStorage and ISynchronizedStorageAdapter implementations, but that is not trivial at the moment, but might be better.

If a single endpoint needs to support multiple tenants using a database-per-tenant architecture, the connection builder needs context information from the incoming message, to read a TenantId or something similar from the message header.

Installers, and the outbox cleaner would not be supported in this scenario without significant changes, and would require knowing all tenant configuration upfront. 

Also, if you are using a transport that doesn't provide native timeout support (Such as SqlTransport), then these will need to be in a shared database (which is already supported) as the timeouts are queried outside of the message context.

Example usage:

Behavior:

```c#
class PropagateIncomingTenantIdBehavior :
	Behavior<ITransportReceiveContext>
{
	public override Task Invoke(ITransportReceiveContext context, Func<Task> next)
	{
	    #region PropagateTenantId
	    
	    if (context.Message.Headers.TryGetValue("TenantId", out var tenant))
	    {
	        context.Extensions.Set("TenantId", tenant);
	    }
	
	    return next();
	
	    #endregion
	}
}
```

Persistence Setup:

```c#
var persistence = endpointConfiguration.UsePersistence<SqlPersistence>();
persistence.ConnectionBuilder(BuildConnection);
persistence.DisableInstaller();
var outbox = endpointConfiguration.EnableOutbox();
outbox.DisableCleanup();
var pipeline = endpointConfiguration.Pipeline;
pipeline.Register(typeof(PropagateIncomingTenantIdBehavior), "Sets the tenant header on the context bag.");
```

Connection Builder:
```c#
private static DbConnection BuildConnection(ContextBag context)
{
    if (context.TryGet<string>("TenantId", out var tenantId))
    {
         //Do Something here
    } 
    return new SqlConnection("connection");
}
```

### DIFF
--- a/src/MsSqlAcceptanceTests.SqlTransport/ConfigureEndpointSqlPersistence.cs
+++ b/src/MsSqlAcceptanceTests.SqlTransport/ConfigureEndpointSqlPersistence.cs
@@ -18,7 +18,7 @@ public class ConfigureEndpointSqlPersistence : IConfigureEndpointTestExecution
         var tablePrefix = TableNameCleaner.Clean(endpointName);
         endpointHelper = new ConfigureEndpointHelper(configuration, tablePrefix, MsSqlConnectionBuilder.Build, BuildSqlDialect.MsSqlServer, FilterTableExists);
         var persistence = configuration.UsePersistence<SqlPersistence>();
-        persistence.ConnectionBuilder(MsSqlConnectionBuilder.Build);
+        persistence.ConnectionBuilder(context => MsSqlConnectionBuilder.Build());
         persistence.SqlDialect<SqlDialect.MsSqlServer>();
         var subscriptions = persistence.SubscriptionSettings();
         subscriptions.DisableCache();
--- a/src/MsSqlAcceptanceTests.SqlTransport/ConfigureEndpointSqlPersistence.cs
+++ b/src/MsSqlAcceptanceTests.SqlTransport/ConfigureEndpointSqlPersistence.cs
@@ -18,7 +18,7 @@ public class ConfigureEndpointSqlPersistence : IConfigureEndpointTestExecution
         var tablePrefix = TableNameCleaner.Clean(endpointName);
         endpointHelper = new ConfigureEndpointHelper(configuration, tablePrefix, MsSqlConnectionBuilder.Build, BuildSqlDialect.MsSqlServer, FilterTableExists);
         var persistence = configuration.UsePersistence<SqlPersistence>();
-        persistence.ConnectionBuilder(context => MsSqlConnectionBuilder.Build());
+        persistence.ConnectionBuilder(MsSqlConnectionBuilder.Build);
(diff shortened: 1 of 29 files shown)


DavidBoike:
Hey Mark! Thanks for the PR!

We're definitely interested in better supporting multi-tenant scenarios. I've raised this PR as a point of discussion for the group responsible.

If you're curious why the test checks failed, it's because of the public API approver. Looks like perhaps quite a few public APIs had to change in order to make this work? Starting immediately, I see, with the ConnectionBuilder to add the ContextBag to the connectionBuilder Func. In order to get that to pass, you have to run that test, and then commit the results of the new API with the code. But then we see the change to APIApprovals.txt and know whether by SemVer standards it could be shipped as a patch, minor, or major release.

I can't make any promises about how quickly we can deal with this, but with breaking changes it could only be a new major release, which would make it even more difficult to ship. Since I haven't dug into it at length, do you think it's possible to release it as a minor, meaning only additive APIs, where parameterless connectionBuilders would, for example, still work? (Not that I want you to rush into trying to make that work, just want to get a gut feeling from you.)

markgould:
The ConnectionBuilder can still take a Func without the ContextBag - so that shouldn't be a breaking change. I just pushed another commit to remove the breaking changes from the ScriptRunner, since that won't have a context anyways. The PR doesn't appear to have updated yet though.

markgould:
@ramonsmits which function is breaking? I added a second function (See SqlPersistenceConfig_Connection.cs) 

The breaking changes are all in non-public classes/methods

ramonsmits:
@markgould You are right, I didn't inspect the changes thoroughly.  I was a change on a public method of a private class that made me write that comment. I should have expanded the whole change.

DavidBoike:
Hey @markgould, we're tackling this now. I'm closing your PR, but you can track what we're doing in https://github.com/Particular/NServiceBus.Persistence.Sql/pull/337.

We're not going to have the connection builder dependent upon the ContextBag, but instead request the header name that will be used to identify the tenant, and allow the ConnectionBuilder to build a connection based on that.

To deal with the outbox cleanup issue, we're going to expose an API that basically says don't clean up the Outbox, I will do that myself via SQL Agent, which is really more of a natural fit for that kind of task anyway. If you opt in to multi-tenant mode, we'll require this option, throwing an exception if you don't.

But going with the "provide the header" route also gives us the option, in the future, to create an in-process cleanup task that tracks tenant ids being used, and is then able to create connections for that outside the message processing pipeline when there is no ContextBag available.

DavidBoike:
@markgould check out https://discuss.particular.net/t/nservicebus-persistence-sql-4-5-1-minor-release-available/1038

markgould:
Excellent! I'll look at what's involved in upgrading our endpoints soon, thanks!

PR #331, opened 2019-01-14
Update development tooling

bording:
This updates development tooling on the `support-3.2` branch.

Note: It is okay for this to be squashed.

@DavidBoike It would be a good idea to switch to Particular.Approvals as part of this PR, but the scope of that work is too much for me to handle as part of all the tooling PRs I'm doing. Do you think you could handle that?


bording:
The compile error is because of the GitVersion bug that chooses the wrong version on the PR branch. :cry:

DavidBoike:
Added support for Particular.Approvals and pushed a temporary `3.2.1-pullrequest1` tag to bypass the GitVersion bug. Now we'll see how the build and tests go.

PR #337, opened 2019-02-15
Multi-tenancy

DavidBoike:


SzymonPobiega:
I like it. The only thing I am not sure is the shape of the API. If I understand correctly, now you either call `persistence.ConnectionBuilder()` or `persistence.MultiTenantConnectionBuilder()`. If you use the latter, you need to be prepared to get `null` as tenant ID and then default to the shared database.

The ideal API from readability point of view would be IMO:

```
persistence.DalayedDelivery.ConnectionBuilder();
persistence.Subscriptions.ConnectionBuilder();
persistence.SynchronizedSession.ConnectionBuilder(); //tenant-enabled Func here
//or
persistence.SynchronizedSession.UseTransportConnection(); //for SQLT
```

The problem is that for simple usages you now need to call three APIs instead of just one. On the other hand, given that only MSMQ requires delayed delivery, a significant portion of the customers would only call two and once SQLT gets native pub/sub, you would just call the `persistence.SynchronizedSession.ConnectionBuilder` in either regular or multi-tenant way.

DavidBoike:
Lol, I essentially had this conversation as a PoA item.

While I like your ideal API from an explicitness perspective I dislike it from the noise factor and obviously because it's breaking.

I think the options are:

1. User needs to be able to supply a connection for a null tenant. I dislike this because it bakes in a lot of assumptions as gotchas.
1. The MultiTenant API registers connection managers for Sagas and Outbox only. On Rabbit, let's say, that's all you'd ever need to do. Try it on MSMQ and it would complain that you don't have a general connection builder for subscriptions and/or timeouts, if those features were enabled.

Obviously I prefer *#*2. I'm not sure I can wrap my head around how it works right now with using the transport connection on SqlT or if you could even use multi-tenant with that.

DavidBoike:
@SzymonPobiega @mauroservienti 		lifting WIP, I think this is ready for your review.

I changed things so if you use MultiTenant connection builder, but you need timeouts or subscriptions, you need to supply a normal ConnectionBuilder too, including a descriptive exception. Also refactored the ConnectionManager into (Multi/Single)TenantConnectionManager with certain methods for invalid cases throwing NotImplementedExceptions, which I think provides a lot more safety, and allowed me to simplify the (internal) API quite a bit to just the 3 operations called by other places in the code.

src/SqlPersistence/Config/ConnectionManager/ConnectionManager.cs
```diff
@@ -0,0 +1,24 @@
...
+    protected async Task<DbConnection> OpenConnection(DbConnection connection)
+    {
+        try
+        {
+            await connection.OpenAsync().ConfigureAwait(false);
+            return connection;
+        }
+        catch
+        {
+            connection.Dispose();
```
mauroservienti:
```suggestion
            connection?.Dispose();
```

Otherwise if connection is null the catch will throw

src/SqlPersistence/Config/ConnectionManager/SingleTenantConnectionManager.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Data.Common;
+using System.Threading.Tasks;
+
+class SingleTenantConnectionManager : ConnectionManager
```
mauroservienti: Is the SingleTenantConnectionManager the one used regularly in non multi-tenant environments? If yes class name is misleading (at least at a first look)

DavidBoike: Yes. At one point I had `MultiTenantConnectionManager : AbstractConnectionManager` and `ConnectionManager : AbstractConnectionManager` but then having `AbstractConnectionManager`s littered all throughout the consuming code just looked yucky.

src/SqlPersistence/Config/SqlPersistenceConfig_Connection.cs
```diff
@@ -15,24 +16,70 @@ public static void ConnectionBuilder(this PersistenceExtensions<SqlPersistence>
...
+                .Set("SqlPersistence.ConnectionManager", new SingleTenantConnectionManager(connectionBuilder));
+        }
+
+        /// <summary>
+        /// Configures how <see cref="DbConnection"/>s are constructed, allowing for selecting a different database per tenant in a multi-tenant system.
+        /// </summary>
+        /// <param name="configuration"></param>
+        /// <param name="tenantIdHeaderName"></param>
+        /// <param name="buildConnectionFromTenantData"></param>
+        public static void MultiTenantConnectionBuilder(this PersistenceExtensions<SqlPersistence> configuration, string tenantIdHeaderName, Func<string, DbConnection> buildConnectionFromTenantData)
```
mauroservienti: My only concern with the configuration API here is: what if tenant information are a runtime evaluated thing based on more than one header? Or you want to transition for a header name to another one?

SzymonPobiega: @mauroservienti in that case you use the second overload (below)

mauroservienti: Right now is private

src/SqlPersistence/Extensions.cs
```diff
@@ -119,4 +109,35 @@ public static bool IsSubclassOfRawGeneric(this Type toCheck, Type generic)
         }
         return false;
     }
+
+    public static IReadOnlyDictionary<string, string> GetMessageHeaders(this ContextBag context)
```
SzymonPobiega: The `TransportReceiveContext` ctor adds the `IncomingMessage` instance to its bag so it is accessible by key in the entire pipeline. I think this could be replaced by just `context.Get<IncomingMessage>().Headers`.

src/SqlPersistence/Config/ConnectionManager/ConnectionManager.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Data.Common;
+using System.Threading.Tasks;
+
+abstract class ConnectionManager
+{
+    public abstract Task<DbConnection> OpenNonContextualConnection();
+    public abstract DbConnection BuildNonContextual();
+    public abstract Task<DbConnection> OpenConnection(IReadOnlyDictionary<string, string> messageHeaders);
+
+    protected async Task<DbConnection> OpenConnection(DbConnection connection)
```
SzymonPobiega: I think if we make this an extension method than we can remove `OpenNonContextualConnection` and only focus in this class on creating connections while the caller would open it via that extension method?

src/SqlPersistence/Config/SqlPersistenceConfig_Connection.cs
```diff
@@ -15,24 +16,70 @@ public static void ConnectionBuilder(this PersistenceExtensions<SqlPersistence>
...
+            if (settings.EndpointIsMultiTenant())
+            {
+                exceptionMessage += $" When in multi-tenant mode with `{nameof(MultiTenantConnectionBuilder)}`, you must still use `{nameof(ConnectionBuilder)}` to provide a database connection for subscriptions/timeouts on message transports that don't support those features natively.";
+            }
+
+            throw new Exception(exceptionMessage);
         }
 
-        internal static Func<DbConnection> GetConnectionBuilder<T>(this ReadOnlySettings settings)
+        // TODO: RENAME to GetConnectionManager
```
SzymonPobiega: I think that given the fact that the connection manager only creates/opens the connection, the name `ConnectionBuilder` would be more appropriate.

src/SqlPersistence/Config/SqlPersistenceConfig_Connection.cs
SzymonPobiega: @mauroservienti right. I didn't notice that.

DavidBoike: With the new refactorings, this is no longer a problem.

DavidBoike:
> The `TransportReceiveContext` ctor adds the `IncomingMessage` instance to its bag so it is accessible by key in the entire pipeline. I think this could be replaced by just `context.Get<IncomingMessage>().Headers`.

This was the linchpin. Instead of focusing on the Headers dictionary, I refactored around IncomingMessage since it's available everywhere. That makes `Func<IncomingMessage, string> captureTenantId` possible, which I think is much more palatable, and needing it for the use case of transitioning from one header to another makes a lot of sense.

Addressed the other comments, that allowed me to simplify the connection manager, make it an interface, and rename SingleTenantConnectionManager to just ConnectionManager.

Please take another look!

src/SqlPersistence/Config/SqlPersistenceConfig_Connection.cs
```diff
@@ -25,13 +26,13 @@ public static void ConnectionBuilder(this PersistenceExtensions<SqlPersistence>
         /// Configures how <see cref="DbConnection"/>s are constructed, allowing for selecting a different database per tenant in a multi-tenant system.
         /// </summary>
         /// <param name="configuration"></param>
-        /// <param name="tenantIdHeaderName"></param>
-        /// <param name="buildConnectionFromTenantData"></param>
+        /// <param name="tenantIdHeaderName">The name of the message header that identifies the tenant id in each message.</param>
+        /// <param name="buildConnectionFromTenantData">Using a tenant id, builds a database connection for that tenant database.</param>
         public static void MultiTenantConnectionBuilder(this PersistenceExtensions<SqlPersistence> configuration, string tenantIdHeaderName, Func<string, DbConnection> buildConnectionFromTenantData)
```
mauroservienti: Is this "shortcut" needed now that we have the one with the `IncomingMessage`?

DavidBoike: I think it's the one that best leads to the pit of success, especially since dealing with the headers properly involves a TryGet. Even if you end up using the more complex one, having this overload tells you "I should be looking at headers for this, not trying to deserialize the message bytes or something."

One thing this brings up in my mind though is that we don't do anything if the multi-tenant builder gets a message with no tenant. It will either throw in the "identify header" stage or pass a null to the build connection stage. Perhaps we should throw a helpful exception? Because none of this is useful without a behavior to keep the tenant ids moving through message flows.

mauroservienti: I think an helpful exception is required.

> Because none of this is useful without a behavior to keep the tenant ids moving through message flows.

Is persistence responsibility to add such a behavior to the pipeline? I guess so. In which case the captured tenant id (and the headers) should be propagated to outgoing messages.

DavidBoike: I don't think it _can_ be the responsibility of the persistence.

If you use the `Func<IncomingMessage, string>` then how do you know which header to propagate into the next message? Your behavior would have to propagate any header that was touched and propagate them all.

To have that kind of info you'd need to remove the `Func<IncomingMessage, string>` option and have something more like `MultiTenantConnectionBuilder(string[] possibleHeadersContainingTenantId, Func<string, DbConnection> buildFromTenantId)`.

If you don't have that all the persistence can do is throw if it is expecting a tenant id and can't find one, prompting the developer to create (with guidance doco) their own behavior. We are in an advanced scenario here, after all.

SzymonPobiega:
What is left here to do?

Issue #340, opened 2019-02-28
When running integration tests with xunit on netcoreapp ScriptLocation returns an invalid path


DavidBoike:
Replaces #312

which means scripts dont get executed

`file:///C:/Users/simon/.nuget/packages/microsoft.testplatform.testhost/15.0.0/lib/netstandard1.5/testhost.dll`

![image](https://user-images.githubusercontent.com/122666/46011482-503f6b00-c109-11e8-9e3a-200bb1ac1caf.png)

seems `Assembly.GetEntryAssembly()` returns the testhost location. which is invalid. While the `AppDomain.CurrentDomain.BaseDirectory` is correct but is doesnt get inside the if

workaround: run them manually

```

    static Task RunSqlScripts(string endpointName, Func<DbConnection> connectionBuilder)
    {
        var baseDir = AppDomain.CurrentDomain.BaseDirectory;
        var scriptDir = Path.Combine(baseDir, "NServiceBus.Persistence.Sql", "MsSqlServer");

        return ScriptRunner.Install(
                sqlDialect: new SqlDialect.MsSqlServer(),
                tablePrefix: endpointName+"_",
                connectionBuilder: connectionBuilder,
                scriptDirectory: scriptDir,
                shouldInstallOutbox: false,
                shouldInstallSagas: true,
                shouldInstallSubscriptions: false,
                shouldInstallTimeouts: false);
    }
```

mikesigs:
I tried the work-around above, but it doesn't really resolve my issue. It does in fact run through the scripts, but it still fails later in the test.

Our integration test is starting the endpoint, without any test-specific modifications to the EndpointConfiguration. Our endpoint is configured with `.EnableInstallers()`. So when it hits the `RunInstallers` step during startup it tries to run the SQL scripts. It fails with a very similar error message to this Issue:

`Expected 'C:\Users\mikes\.nuget\packages\microsoft.testplatform.testhost\15.9.0\lib\netstandard1.5\NServiceBus.Persistence.Sql\MsSqlServer\Outbox_Create.sql' to exist. It is possible it was not deployed with the endpoint.`

I'll keep digging through this on my own, but was hoping there might be some progress on permanent fix for this issue.

Thanks!

DavidBoike:
@mikesigs we're getting ready to release the fix to this. Stay tuned.

InLoop-TOCA:
Hi @DavidBoike 

[...]

PR #351, opened 2019-07-09
Use the same location detection as core

WilliamBZA:
Fixes https://github.com/Particular/NServiceBus.Persistence.Sql/issues/340

This changes the current directory code to match the same code that is used in [Core](https://github.com/Particular/NServiceBus/blob/65fee422bdd79a3e5d4adc3dcc40353cb5b9dd82/src/NServiceBus.Core/Hosting/Helpers/AssemblyScanner.cs#L21)

I'm not certain of the underlying implications here. From looking at the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.appdomain.basedirectory) for the property, this should give the correct location for all framework and netstandard types that we support.

This means that using the `AppDomain.CurrentDomain.BaseDirectory` should be the right approach.

I am not sure what cases the `Assembly.GetEntryAssembly()` codepath was trying to solve that weren't solved by `AppDomain.CurrentDomain.BaseDirectory`. In my tests it all seemed to work.

### DIFF
--- a/src/SqlPersistence/ScriptLocation.cs
+++ b/src/SqlPersistence/ScriptLocation.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.IO;
-using System.Reflection;
 using NServiceBus;
 using NServiceBus.Settings;
 
@@ -10,30 +9,20 @@ static class ScriptLocation
 
     public static string FindScriptDirectory(ReadOnlySettings settings)
     {
-        var currentDirectory = GetCurrentDirectory(settings);
+        var currentDirectory = GetScriptsRootPath(settings);
         return Path.Combine(currentDirectory, ScriptFolder, settings.GetSqlDialect().Name);
     }
 
-    static string GetCurrentDirectory(ReadOnlySettings settings)
+    static string GetScriptsRootPath(ReadOnlySettings settings)
     {
         if (settings.TryGet("SqlPersistence.ScriptDirectory", out string scriptDirectory))
         {
             return scriptDirectory;
         }
-        var entryAssembly = Assembly.GetEntryAssembly();
-        if (entryAssembly == null)
-        {
-            var baseDir = AppDomain.CurrentDomain.BaseDirectory;
-            var scriptDir = Path.Combine(baseDir, ScriptFolder);
-            //if the app domain base dir contains the scripts folder, return it. Otherwise add "bin" to the base dir so that web apps work correctly
-            if (Directory.Exists(scriptDir))
-            {
-                return baseDir;
-            }
-            return Path.Combine(baseDir, "bin");
-        }
-        var codeBase = entryAssembly.CodeBase;
-        return Directory.GetParent(new Uri(codeBase).LocalPath).FullName;
+
+        //NOTE: This is the same logic that Core uses for finding assembly scanning path.
+        //      RelativeSearchPath is set for ASP .NET and points to the binaries folder i.e. /bin
+        return AppDomain.CurrentDomain.RelativeSearchPath ?? AppDomain.CurrentDomain.BaseDirectory;
     }
 
     public static void ValidateScriptExists(string createScript)


src/SqlPersistence/ScriptLocation.cs
```diff
@@ -20,20 +19,15 @@ static string GetCurrentDirectory(ReadOnlySettings settings)
         {
             return scriptDirectory;
         }
-        var entryAssembly = Assembly.GetEntryAssembly();
```
bording: I have no idea why `GetEntryAssembly` was being used. I would have expected `GetExecutingAssembly` instead.

src/SqlPersistence/ScriptLocation.cs
```diff
@@ -20,20 +19,15 @@ static string GetCurrentDirectory(ReadOnlySettings settings)
...
-        if (entryAssembly == null)
+
+        var baseDir = AppDomain.CurrentDomain.BaseDirectory;
```
bording: So this change likely works, but it's still relying on something that can have some edge cases. `AppDomain.CurrentDomain.BaseDirectory` is set by the thing the created the `AppDomain`, so we're relying on the fact that it's been set correctly in all scenarios now. Probably safe generally, but it's certainly possible to make it fail.

src/SqlPersistence/ScriptLocation.cs
```diff
@@ -20,20 +19,15 @@ static string GetCurrentDirectory(ReadOnlySettings settings)
...
-            if (Directory.Exists(scriptDir))
-            {
-                return baseDir;
-            }
-            return Path.Combine(baseDir, "bin");
+            return baseDir;
         }
-        var codeBase = entryAssembly.CodeBase;
-        return Directory.GetParent(new Uri(codeBase).LocalPath).FullName;
+        return Path.Combine(baseDir, "bin");
```
bording: Appending `bin` to the path without checking it exists and contains the scripts folder is assuming that the only reason that the `if` failed is because we're running inside ASP.NET. I'm wondering if this should be more defensive.

src/SqlPersistence/ScriptLocation.cs
```diff
@@ -20,20 +19,15 @@ static string GetCurrentDirectory(ReadOnlySettings settings)
...
-            //if the app domain base dir contains the scripts folder, return it. Otherwise add "bin" to the base dir so that web apps work correctly
-            if (Directory.Exists(scriptDir))
-            {
-                return baseDir;
-            }
-            return Path.Combine(baseDir, "bin");
+            return baseDir;
         }
-        var codeBase = entryAssembly.CodeBase;
-        return Directory.GetParent(new Uri(codeBase).LocalPath).FullName;
```
bording: I'd almost consider suggesting switching `GetEntryAssembly` to `GetExecutingAssembly` and leaving the rest of the code as-is, but this line also isn't quite correct. The `CodeBase` property might look like a uri, but it's not actually valid, so passing it to the `Uri` constructor can fail in some cases.

And you can't rely on `Location` instead because that will reflect any shadow copying that could be going on, which would be the wrong path.

WilliamBZA:
After chatting to @tmasternak, we feel that this change is a sane, safe one to make. As long as we use the same method for finding the directory that core does, the behaviour is at least predictable. If `AppDomain.BaseDirectory` behaves differently for some reason in a future environment, then the assembly scanning will also behave differently.

Further, since the script generation is likely to be a development time concern, the workarounds (specify the directory manually) means that this still won't be a blocker.

bording:
@WilliamBZA Take a look at https://github.com/Particular/NServiceBus/blob/0dfe20bf5a872b491bbaafdf7e0cb9d23580710d/src/NServiceBus.Core/EndpointConfiguration.cs#L136

This is what Core actually does. Currently when running under ASP.NET, `RelativeSearchPath` is the full path to the bin folder. If you do the same thing here, you can remove even more code.

tmasternak:
@bording I've switched to the Core's logic for finding assembly scanning path.

tmasternak:
@WilliamBZA It looks that the integration tests are failing. Can this be somehow connected to the other PR for generating case sensitive collation scripts?

src/SqlPersistence/ScriptLocation.cs
```diff
@@ -10,30 +9,20 @@ static class ScriptLocation
...
-            {
-                return baseDir;
-            }
-            return Path.Combine(baseDir, "bin");
-        }
-        var codeBase = entryAssembly.CodeBase;
-        return Directory.GetParent(new Uri(codeBase).LocalPath).FullName;
+
+        //NOTE: This is the same logic that Core uses for finding assembly scanning path.
+        //      RelativeSearchPath is set for ASP .NET and points to the binaries folder i.e. /bin
```
bording: Just an FYI, but despite the name, `RelativeSearchPath` contains an absolute path to the bin folder when running in ASP.NET

Issue #347, opened 2019-03-28
Script generation fails on case sensitive coalitions with SQL Server


danielmarbach:
## Who's affected

All customers using case-sensitive coalitions

## Symptoms

When using case sensitive coalitions the script execution fails with the following exception

```
Invalid object name ‘information_schema.columns’
```

## Workaround

Generate the scripts once, disable script generation and then modify the mentioned queries against `information_schema.columns` by using the proper casing for both the table name and the columns

## Analysis

Case sensitive coalitions don't know what `information_schema.columns` is because the database table is actually named `INFORMATION_SCHEMA.COLUMNS`. 

The script generation needs to be adjusted 

https://github.com/Particular/NServiceBus.Persistence.Sql/blob/e08295d842e74996127d55cc6f86d79cf07f8a09/src/ScriptBuilder/Saga/Writers/MsSqlServerSagaScriptWriter.cs#L70

https://github.com/Particular/NServiceBus.Persistence.Sql/blob/e08295d842e74996127d55cc6f86d79cf07f8a09/src/ScriptBuilder/Saga/Writers/MsSqlServerSagaScriptWriter.cs#L125

For case insensitive coalitions `information_schema.columns` and `INFORMATION_SCHEMA.COLUMNS` are equivalent and thus the capital cased one should be used as a default.

```
select * from information_schema.columns
```

![image](https://user-images.githubusercontent.com/174258/55157538-f3c31400-515c-11e9-9742-c4e942a703ed.png)

vs

![image](https://user-images.githubusercontent.com/174258/55157565-02a9c680-515d-11e9-9aef-1bf9f8a0bdfa.png)

or the proof that they return the same result

```
select * from information_schema.columns
except
select * from INFORMATION_SCHEMA.COLUMNS
```

![image](https://user-images.githubusercontent.com/174258/55157592-18b78700-515d-11e9-9a1a-ebe3c1595cb4.png)

Given that the column names are capital cased as well it makes sense to also capital case the column names to make it properly work

Taking a quick glance it looks like other tables like `sys.indexes` are not affected. It seems like it really only affects `information_schema.columns`

As an example

```
from information_schema.columns
    where
        table_name = @tableNameWithoutSchema and
        table_schema = @schema and
        column_name like 'Correlation_%'{builder}
```

should be changed to 

```
from INFORMATION_SCHEMA.COLUMNS
    where
        TABLE_NAME = @tableNameWithoutSchema and
[...]

PR #352, opened 2019-07-15
Change information_schema to uppercase

WilliamBZA:
This change should also include a change to the Sql Server used in the test environment to be case sensitive

* [x] Change TC server to be case sensitive

Fixes https://github.com/Particular/NServiceBus.Persistence.Sql/issues/347

### DIFF
--- a/src/TestHelper/MsSqlConnectionBuilder.cs
+++ b/src/TestHelper/MsSqlConnectionBuilder.cs
@@ -8,10 +8,14 @@ public static class MsSqlConnectionBuilder
     public static SqlConnection Build()
     {
         var connection = Environment.GetEnvironmentVariable("SQLServerConnectionString");
+
         if (string.IsNullOrWhiteSpace(connection))
         {
-            return new SqlConnection(ConnectionString);
+            SetCollationToCaseSensitive(ConnectionString);
+            return new SqlConnection(connection);
         }
+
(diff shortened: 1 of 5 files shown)


WilliamBZA:
Case sensitivity for sql server is now tested during the test phase

src/TestHelper/MsSqlConnectionBuilder.cs
```diff
@@ -33,7 +33,7 @@ public static void Setup(string tenantId)
             using (var conn = MsSqlConnectionBuilder.Build())
             {
                 conn.Open();
-                conn.ExecuteCommand($"if not exists (select * from sysdatabases where name = '{dbName}') create database {dbName};");
+                conn.ExecuteCommand($"if not exists (select * from sysdatabases where name = '{dbName}') create database {dbName}; ALTER DATABASE [{dbName}] COLLATE SQL_Latin1_General_CP1_CS_AS;");
```
tmasternak: What happens on the build agent when the db already exists?

WilliamBZA: The command just executes.

WilliamBZA: I take that back, it only executes if the schema is already case sensitive...

WilliamBZA:
Closing this as I want to take a different tack on this one.

PR #353, opened 2019-07-17
Check for collation correctness

### LINKED ISSUE #347: Script generation fails on case sensitive coalitions with SQL Server

WilliamBZA:
* Fixes #347 by changing information_schema.columns to uppercase
* Changes the tests to make sure the database is in a case sensitive collation
  * First checks if the exists but has incorrect collation. If yes, then it drops the db
  * Next checks if DB exists. If no, creates db with correct collation

### DIFF
--- a/src/MsSqlAcceptanceTests/SetUpFixture.cs
+++ b/src/MsSqlAcceptanceTests/SetUpFixture.cs
@@ -1,0 +1,12 @@
+using NUnit.Framework;
+
+[SetUpFixture]
+public class SetUpFixture
+{
+    [OneTimeSetUp]
+    public void SetUp()
+    {
+        MsSqlConnectionBuilder.DropDbIfCollationIncorrect();
+        MsSqlConnectionBuilder.CreateDbIfNotExists();
+    }
+}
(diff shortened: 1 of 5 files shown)


WilliamBZA:
...Yes. The tests won't pass on old release branches unless this is backported

PR #357, opened 2019-09-20
Use interface for path that is used outside the customer saga finders

danielmarbach:
This uses the interface that is already public for the path that is used in the handler to enable simple testing over the testable context. It doesn't solve the problem for the saga finder but at least gets customer unstuck in a simple way for the handlers. I don't think the custom finders are a problem given that we don't recommend to use them anyway

### DIFF
--- a/src/SqlPersistence/SqlPersistenceStorageSessionExtensions.cs
+++ b/src/SqlPersistence/SqlPersistenceStorageSessionExtensions.cs
@@ -22,7 +22,16 @@
             return GetSqlStorageSession(session);
         }
 
-        static StorageSession GetSqlStorageSession(this SynchronizedStorageSession session)
+        static ISqlStorageSession GetSqlStorageSession(this SynchronizedStorageSession session)
+        {
+            if (session is ISqlStorageSession storageSession)
+            {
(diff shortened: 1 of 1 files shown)


src/SqlPersistence/SqlPersistenceStorageSessionExtensions.cs
```diff
@@ -22,7 +22,16 @@ public static ISqlStorageSession SqlPersistenceSession(this SynchronizedStorageS
             return GetSqlStorageSession(session);
         }
 
-        static StorageSession GetSqlStorageSession(this SynchronizedStorageSession session)
+        static ISqlStorageSession GetSqlStorageSession(this SynchronizedStorageSession session)
```
timbussmann: couldn't this just be inlined at line 22?

danielmarbach: Done

mauroservienti:
@danielmarbach can you please rebased this on master? It should fix the inspection error(s)

danielmarbach:
@mauroservienti done

PR #364, opened 2019-10-11
Put targets file in buildTransitive also

bording:
NuGet 5.0 introduced support for a `buildTransitive` folder that ensures that any props or targets files in it are transitively referenced across project references.

Add the targets file to this folder means that any sagas that are in a project that only has a transitive reference to the persistence package will also have their scripts created.

This means that anyone using NuGet 5.0, for example from VS 2019 or the .NET Core 3.0 SDK, will get the expected behavior.

And if for some reason they don't want the saga scripts to be created, the [already documented](https://docs.particular.net/persistence/sql/controlling-script-generation#disabling-script-generation) way of disabling it still works.


bording:
There was an unrelated build failure related to remaining package wildcard use, so I also got rid of that.

PR #366, opened 2019-10-25
Use GeneratePathProperty to simplify project file

bording:
This gets rid of needing a special way to handle the version of the NServiceBus.AcceptanceTests.Sources package by using the new `GeneratePathProperty` feature.

Unfortunately, we can't merge this yet because it requires VS 2019 to use from inside VS, and our build agents haven't been updated yet.


danielmarbach:
This should be mergeable now right?

PR #368, opened 2019-10-29
Setting LangVersion to 7.3

HEskandari:


HEskandari:
This one had the same issue, plus an additional dependency has become incompatible (since we use wildcard versioning). The build should pass now.

HEskandari:
Closing as the build has stuck.

PR #376, opened 2019-12-11
Adds ROWLOCK hint

ramonsmits:
A customer is running into issues with lock escalation. I was wondering if this could be prevented by adding the ROWLOCK hint.

>ROWLOCK
>Specifies that row locks are taken when page or table locks are ordinarily taken. When specified in transactions operating at the SNAPSHOT isolation level, row locks are not taken unless ROWLOCK is combined with other table hints that require locks, such as UPDLOCK and HOLDLOCK.

https://docs.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-table?view=sql-server-ver15

### DIFF
--- a/src/SqlPersistence/Saga/SqlDialect_MsSqlServer.cs
+++ b/src/SqlPersistence/Saga/SqlDialect_MsSqlServer.cs
@@ -89,7 +89,7 @@ select
     Metadata,
     Data
 from {tableName}
-with (updlock)
+with (updlock,rowlock)
 where Id = @Id
 ";
             }
@@ -104,7 +104,7 @@ select
     Metadata,
     Data
 from {tableName}
-with (updlock)
+with (updlock,rowlock)
 where Correlation_{propertyName} = @propertyValue
 ";
             }
@@ -127,7 +127,7 @@ select
     Metadata,
     Data
 from {tableName}
-with (updlock)
+with (updlock,rowlock)
 where {whereClause}
 ";
             }


SzymonPobiega:
Do we have any evidence that the hint would help? Locking a row should be default. The way NServiceBus runs the saga persistence means that for a single connection there is going to be exactly one row locked so I don't understand why would SQL Server escalate locks.

My understanding of escalation is that SQL Server does it when a single connection holds many row locks in the same table.

danielmarbach:
Interesting read https://littlekendra.com/2016/02/04/why-rowlock-hints-can-make-queries-slower-and-blocking-worse-in-sql-server/

andreasohlund:
Is this still a thing?

danielmarbach:
Based on the article I linked I'd say no

ramonsmits:
The article is talking about thousands of locks within a single table where a SINGLE select has a multi row result set. This is not the case with sagas where there is always just a single row match.

Just look at the types of queries in the article, unbounded result sets. Obviously that will have a very negative impact when enforcing row locking.

The problem is that the CURRENT design results in lock escalation which can be prevented.

The mentioned workaround "Workaround: Use Optimistic Locking (Snapshot or Read Committed Snapshot) for SELECT Queries" can back fire if there is lots of single item contention. And second, that ONLY applies if you actually have two atomic operations not within a single transactions.

The article ends with:

>Takeaways: Use ROWLOCK Hints Cautiously
>I’m not against ROWLOCK hints, I’ve just seen them backfire. If you choose to force row locking as part of your solution, check to make sure it’s not slowing down your queries and keep a close eye on blocking and lock escalation on your tables.


So, reopening as yes, its still a thing.

SzymonPobiega:
@ramonsmits what about my comment above? My understanding is that locks are tied to transactions. In our case we always have a single row accessed in a transaction so I don't understand how lock escalation can happen. There is now way for SQL Server to consolidate locks owned by different connection/transactions into a page or table lock.

ramonsmits:
> @ramonsmits what about my comment above? My understanding is that locks are tied to transactions. In our case we always have a single row accessed in a transaction so I don't understand how lock escalation can happen. There is now way for SQL Server to consolidate locks owned by different connection/transactions into a page or table lock.

@SzymonPobiega AFAIK Lock escalation can happen due to starvation of lock resources. I'm not sure if this will actually resolves the lock escalation of the customer as we could not test this because SQL queries are unfortunately hard coded. I do know that lock escalation happened on the saga table resulting deadlocks and reduced throughput.

I reopened because the lock escalation is not resolved and that this MIGHT be the solution to it.

ramonsmits:
@SzymonPobiega Besides this HINT, there is also https://github.com/Particular/NServiceBus.Persistence.Sql/pull/723 which I would love to see merged.

PR #382, opened 2020-01-03
Fixed InvalidCastException when using Microsoft.Data.SqlClient

robbaman:
Fixes the InvalidCastException that occurs when initializing the NServiceBus Persistance with an SqlConnection instance from the Microsoft.Data.SqlClient package.

See this issue/comment for a description of the problem:
https://github.com/Particular/NServiceBus.SqlServer/issues/506#issuecomment-570193933

### DIFF
--- a/src/SqlPersistence/Config/SqlDialect_MsSqlServer.cs
+++ b/src/SqlPersistence/Config/SqlDialect_MsSqlServer.cs
@@ -34,11 +34,18 @@ namespace NServiceBus
                 //TODO: do ArraySegment fro outbox
                 if (value is ArraySegment<char> charSegment)
                 {
-                    var sqlParameter = (SqlParameter)parameter;
-
-                    sqlParameter.Value = charSegment.Array;
-                    sqlParameter.Offset = charSegment.Offset;
-                    sqlParameter.Size = charSegment.Count;
+                    if (parameter is SqlParameter sqlParameter) {
+                        sqlParameter.Value = charSegment.Array;
+                        sqlParameter.Offset = charSegment.Offset;
+                        sqlParameter.Size = charSegment.Count;
+                    }
+                    #if NETSTANDARD2_0
+                    else if (parameter is Microsoft.Data.SqlClient.SqlParameter msSqlParameter) {
+                        msSqlParameter.Value = charSegment.Array;
+                        msSqlParameter.Offset = charSegment.Offset;
+                        msSqlParameter.Size = charSegment.Count;
+                    }
+                    #endif
                 }
                 else
                 {
--- a/src/SqlPersistence/SqlPersistence.csproj
+++ b/src/SqlPersistence/SqlPersistence.csproj
@@ -16,7 +16,8 @@
     <PackageReference Include="Obsolete.Fody" Version="4.3.2" PrivateAssets="All" />
     <PackageReference Include="Particular.Packaging" Version="0.5.0" PrivateAssets="All" />
     <PackageReference Include="System.Data.SqlClient" Version="4.4.3" />
-  </ItemGroup>
+    <PackageReference Include="Microsoft.Data.SqlClient" Version="1.1.0" Condition="$(TargetFramework)=='netstandard2.0'" />
+ </ItemGroup>
 
   <ItemGroup>
     <Compile Include="..\Guard.cs" />


mauroservienti:
@robbaman thanks for raising this, I already pinged the relevant group to have a look at your proposed changes.

DavidBoike:
Hey @robbaman,

Unfortunately I don't think we can accept this PR. We really don't want to add an additional dependency. What we'd really like to do is to remove the dependency on BOTH SqlClient implementations, using only DbConnection and friends, so that it doesn't matter which you use at runtime.

Of course that introduces the problem of what to do when there is a specific need to do an implementation-specific thing. Do we just break down and use reflection? Then that becomes brittle, so how do we make sure that doesn't/can't break? Plus reflection isn't incredibly fast, if the code is on a hot path we need to be cognizant of that and perhaps do something to mitigate it.

And of course the biggest thing is we'd like to fix the issue with SQL Persistence at the same time as SQL Transport, because we have a lot of customers using both. SQL Persistence is easier because it was built from the ground up to be largely client-independent so that it can also support MySQL/Postgres/Oracle, but in SQL Transport it's not nearly so easy.

So I know it's painful at the moment, but we are working toward it. And knowing that we have a customer formally blocked by it has raised the priority of a fix. So stay tuned.

robbaman:
I understand your reasoning, but what do you suggest we do now? Can you at least produce a version that doesn't crash by removing the cast/dependency altogether or create a 'Do magic stuff to the DbParameter'-abstraction?

We're running in to this problem in any service that uses both Saga's and EntityFrameworkCore 3.0 which is all our new services.

DavidBoike:
@robbaman If the code in your PR works for your situation, you could build that package locally and use it in the interim until we can officially support Microsoft.Data.SqlClient and EFCore 3.0.

robbaman:
> @robbaman If the code in your PR works for your situation, you could build that package locally and use it in the interim until we can officially support Microsoft.Data.SqlClient and EFCore 3.0.

That's what we're currently doing in development, but was hoping we didn't have to resort to this when we go to production as this means we'll have to keep making new packages as updates in your package are available.

Is there any approximate ETA when you expect to support Microsoft.Data.SqlClient for the NServiceBus.Persistence.Sql package? I mean are we talking days/weeks or months/years?

Thanks

DavidBoike:
I can't give you an ETA, but I can say that it's extremely unlikely that there would be an update to SQL Persistence *before* we release support for the new SQL Client, unless some critical/security bug was discovered.

robbaman:
I'm afraid I spoke too soon. It worked on on existing project, but creating a new service with our changed implementation of the service leads to an error in the ScriptBuilder during build:

```
Error	MSB4018	The "SqlPersistenceScriptBuilderTask" task failed unexpectedly.
System.IO.FileNotFoundException: Could not load file or assembly 'NServiceBus.Persistence.Sql.ScriptBuilder, Version=4.0.0.0, Culture=neutral, PublicKeyToken=9fc386479f8a226c' or one of its dependencies. The system cannot find the file specified.
File name: 'NServiceBus.Persistence.Sql.ScriptBuilder, Version=4.0.0.0, Culture=neutral, PublicKeyToken=9fc386479f8a226c'
   at NServiceBus.Persistence.Sql.SqlPersistenceScriptBuilderTask.Execute()
   at Microsoft.Build.BackEnd.TaskExecutionHost.Microsoft.Build.BackEnd.ITaskExecutionHost.Execute()
   at Microsoft.Build.BackEnd.TaskBuilder.<ExecuteInstantiatedTask>d__26.MoveNext()

WRN: Assembly binding logging is turned OFF.
To enable assembly bind failure logging, set the registry value [HKLM\Software\Microsoft\Fusion!EnableLog] (DWORD) to 1.
Note: There is some performance penalty associated with assembly bind failure logging.
To turn this feature off, remove the registry value [HKLM\Software\Microsoft\Fusion!EnableLog].
	JClub.Portal.MessageBus	C:\Users\mulde\.nuget\packages\jclub.nservicebus.persistence.sql\1.1.1-beta2\buildTransitive\JClub.NServiceBus.Persistence.Sql.targets	18	
```

Any ideas where this may come from and how to fix it?

robbaman:
Turns out you were one hack ahead of me with the whole ILRepack thing which only works in release builds.

It's fixed now... nothing to see here...

PR #417, opened 2020-02-14
add 'CreateTime' field for OutBox

Varorbc:
when the table data is big, then delete the data, it doesn't work well.

### DIFF
--- a/src/ScriptBuilder/Outbox/Create_MsSqlServer.sql
+++ b/src/ScriptBuilder/Outbox/Create_MsSqlServer.sql
@@ -12,9 +12,10 @@ set @createTable = '
     create table ' + @tableName + '(
         MessageId nvarchar(200) not null primary key nonclustered,
         Dispatched bit not null default 0,
-        DispatchedAt datetime,
+        DispatchedAt datetime2,
         PersistenceVersion varchar(23) not null,
-        Operations nvarchar(max) not null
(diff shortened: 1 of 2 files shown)


SzymonPobiega:
@Varorbc Unfortunately we can't accept this PR as it would be a breaking change in the schema. Doing such breaking changes requires a migration strategy for the existing data.

If you have issues running the Outbox cleanup, please reach out to us via support channel. One option to make the cleanup faster is to disable the built-in cleanup and define partitions (e.g. weekly) in the outbox table. Then you can drop entire partition as part of a weekly SQL Server Agent job.

ramonsmits:
The suggested behavior would be to remove based on creation time instead of dispatch time which seems incorrect. If dispatched at "t+n" and for some reason the same message would be processed after a few seconds "t+n+5s" (for example, expired lease) and cleanup happened you now are vulnerable to more than once processing.

Varorbc:
> @Varorbc Unfortunately we can't accept this PR as it would be a breaking change in the schema. Doing such breaking changes requires a migration strategy for the existing data.
> 
> If you have issues running the Outbox cleanup, please reach out to us via support channel. One option to make the cleanup faster is to disable the built-in cleanup and define partitions (e.g. weekly) in the outbox table. Then you can drop entire partition as part of a weekly SQL Server Agent job.

Thank you for your suggestion, I will close this pull request

PR #425, opened 2020-02-24
Update test projects to netcoreapp3.1 in release-4.5

bording:
For your review @mauroservienti


mauroservienti:
@bording now that the MySql.Data package dependency is locked, everything is green. Should we also lock the other dependencies that are using wildcards?

https://github.com/Particular/NServiceBus.Persistence.Sql/blob/773ae2846797e3d4c5cc476cf4b16df8c5931055/src/TestHelper/TestHelper.csproj#L9-L17

bording:
> Should we also lock the other dependencies that are using wildcards?

It wouldn't be a bad idea, but I there are likely to be a lot of them.

mauroservienti:
Good point. I'll merge this one as is and raise follow up issues to look for and fix those.

PR #428, opened 2020-03-02
Update test projects to NServiceBus 7.2.3 in release-4.5

bording:
For your review @mauroservienti

PR #429, opened 2020-03-02
Update test projects to NServiceBus 7.2.3 in release-4.6

bording:
For your review @mauroservienti


mauroservienti:
Seems to be the same problem as #428

PR #442, opened 2020-03-31
GitHubSync update

particularbot:
This is an automated synchronization PR.

The following source template repositories were used:
* particular/RepoStandards


.gitignore
```diff
@@ -96,7 +92,4 @@ src/scaffolding.config
 
 # JetBrains Rider
 .idea/
-*.sln.iml
-src/Integration/Directory.Build.props
-
-PromotedSqlScripts/
```
andreasohlund, line 102: This is still needed https://github.com/Particular/NServiceBus.Persistence.Sql/blob/master/src/Integration/Shared/SqlPersistenceSettings.cs#L8 

Do I add a nested .gitignore like we discussed @bording @ramonsmits ?

PR #470, opened 2020-06-10
Use official package for ScriptDom dependency

bording:
This PR switches the ScriptBuilder.Tests project over to using the official Microsoft package to bring in the Microsoft.SqlServer.TransactSql.ScriptDom dependency,

This has a few impacts:
- The official package does bring in more stuff than just ScriptDom, but that shouldn't actually be a problem.
- The official package has a `net46` minumum, so I had to bump the test project up.
- The tests can actually be run on .NET Core now.


bording:
@SzymonPobiega The linux test failure was just a problem with the linux build agents that I fixed by rebooting them last week. I'll go ahead and merge now that this is green.

bording:
Actually, I just noticed some whitespace cleanup I need to do, so I'll do that first.

PR #473, opened 2020-06-18
Synchronized session DI

### LINKED ISSUE #375: Register Synchronized Storage Session in the Dependency Injection container
I've been trying to use SQL persistence and outbox in an existing project, without having to change too much code. One thing I would like to do is to reuse a `SqlConnection` for both the Outbox feature, as well as for dependencies registered in the child container that is created for the incoming message pipeline. 

```csharp
// Instead of writing this:
persistence.ConnectionBuilder(() => new SqlConnection("connection string");

// I would like to do this:
persistence.ConnectionBuilder((IBuilder builder) => builder.Build<SqlConnection>();
```

Right now, if I want to use the same `SqlConnection` for my message pipeline **and** resolving dependencies (like an EF DbContext), I can only do that by using `context.SynchronizedStorageSession.SqlPersistenceSession().Connection` in my message handler, obtaining the existing connection, and then create another child of the child container where I register that connection somehow. **I end up with quite a bit of boiler plate infrastructure code in my message handlers.**

With the method above, I would be able to use constructor injection directly on the message handler. Unfortunately it's exceedingly difficult to achieve this as things are now. `IConnectionManager` is internal, and even if it weren't, I don't see a way I would be able to access the `RootContext` or something that gives me access to the child container. I guess that `RootContext` would have to be part of the `IConnectionManager.Build(...)` method signature.

I've managed a workaround using by combining a callback when the child container gets built, `AsyncLocal<T>` and a service locator pattern inside my ConnectionBuilder Func, but needless to say it's not looking pretty.

SzymonPobiega:
Fixes #375

This change adds registration of `ISqlStorageSession` in the DI container so that the connection and transaction it contains can be injected into user code (e.g. repositories).

Following code from `When_using_outbox_synchronized_session_via_container` demonstrates this usage:

```
cc.ConfigureComponent<MyRepository>(DependencyLifecycle.InstancePerUnitOfWork);
cc.ConfigureComponent(b => b.Build<ISqlStorageSession>().Connection, DependencyLifecycle.InstancePerUnitOfWork);
```

Based on the [design document](https://github.com/Particular/DeveloperExperience/blob/master/research/synchronized-session/index.md) we don't want to register the `DbConnection` itself in the container because that is a common type and some users might already have it registered using different approaches.

Contrary to what has been initially written in that document, we use `ISqlStorageSession` here instead of a class to make the API as close to the current, context-based one, as possible. If we did use a class (e.g. `SqlStorageSession`) it could be confusing for the users why they need to use different types depending on if they use the session via DI or via message handling conetxt.

NOTE: This PR also bumps the framework version from 452 to 461. Before merging it we need to consider if we want to:
 - bump the fx version in minor
 - bump the fx version in major (and release major)
 - use shim and revert back to 452


src/SqlPersistence/SynchronizedStorage/StorageSessionFeature.cs
```diff
@@ -95,4 +98,6 @@ static JsonSerializer BuildJsonSerializer(JsonSerializerSettings jsonSerializerS
         }
         return JsonSerializer.Create(jsonSerializerSettings);
     }
+
+    CurrentSessionHolder sessionHolder = new CurrentSessionHolder();
```
tmasternak, line 102: Do we need this field? Can this be a variable in the `Setup` method?

PR #495, opened 2020-07-27
Persistence tests for all 4 dialects

SzymonPobiega:


src/SqlPersistence.PersistenceTests/SqlPersistence.PersistenceTests.csproj
```diff
@@ -0,0 +1,32 @@
...
+    <PackageReference Include="NUnit3TestAdapter" Version="3.17.0" />
+    <PackageReference Include="Particular.Approvals" Version="0.2.0" />
+    <PackageReference Include="PublicApiGenerator" Version="9.3.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <!---  Remove this exclusion once PR is merged: https://github.com/Particular/NServiceBus/pull/5707 and remove When_worker_tries_to_complete_saga_update_by_another_pessimistic_tweaked.cs-->
+    <Compile Remove="\**\When_worker_tries_to_complete_saga_update_by_another_pessimistic.cs" />
+
+    <Compile Remove="\**\When_concurrent_update_exceed_lock_request_timeout_pessimistic.cs" />
```
bording, line 27: Do we need a comment about why this one is excluded?

src/SqlPersistence.PersistenceTests/PersistenceTestsConfiguration.cs
```diff
@@ -0,0 +1,171 @@
...
+    using NServiceBus.Outbox;
+    using NServiceBus.Sagas;
+    using NUnit.Framework;
+    using Persistence;
+    using Persistence.Sql.ScriptBuilder;
+    using Transport;
+
+    public partial class PersistenceTestsConfiguration
+    {
+        public bool SupportsDtc => false; // TODO: this should actually be set to true, tests still failing
```
bording, line 20: What do we need to do with this TODO?

src/SqlPersistence.PersistenceTests/PersistenceTestsConfiguration.cs
```diff
@@ -0,0 +1,171 @@
...
+            using (var connection = connectionFactory())
+            {
+                connection.Open();
+
+                foreach (var saga in SagaMetadataCollection)
+                {
+                    CorrelationProperty correlationProperty = null;
+                    if (saga.TryGetCorrelationProperty(out var propertyMetadata))
+                    {
+                        //TODO: Hard-code correlation property to string
```
bording, line 122: Here's another TODO. What do we need to do here?

src/SqlPersistence.PersistenceTests/SqlPersistence.PersistenceTests.csproj
lailabougria: They are all documented in the issue, the only reason that I added a comment to the other one, is because that's actually already fixed, the package is just not yet available

src/SqlPersistence.PersistenceTests/PersistenceTestsConfiguration.cs
lailabougria: Actually, now that I see it, I need to add this to the ignored tests issue and the todo can be removed

src/SqlPersistence.PersistenceTests/PersistenceTestsConfiguration.cs
bording: Going to go ahead and remove it.

PR #506, opened 2020-08-20
Introduce proper disposal of OutboxTransaction in cause it faulted during begin

### LINKED ISSUE #504: The new 'UseTransactionScope' option for outbox does not work as expected
We are running NServicebus 7.4.0 with `NServiceBus.Persistence.Sql`  6.0.2. We have recently activated `TransactionScope` as the transaction type for the outbox according [to the NServiceBus documentation](https://docs.particular.net/persistence/sql/outbox#transaction-type-transaction-scope). 

To my surprise, `Transaction.Current` is `null` in my message handler. Here's a sample app that can be used to reproduce the issue: https://github.com/pardahlman/nsb-trasaction-scope-repro

Have I missed something, or is this expected behavior?

danielmarbach:
Fixes #504

### DIFF
--- a/src/SqlPersistence/Outbox/OutboxPersister.cs
+++ b/src/SqlPersistence/Outbox/OutboxPersister.cs
@@ -33,7 +33,24 @@ class OutboxPersister : IOutboxStorage
         var transaction = outboxTransactionFactory();
         transaction.Prepare(context);
         // we always need to avoid using async/await in here so that the transaction scope can float!
-        return transaction.Begin(context);
+        return BeginTransactionInternal(transaction, context);
+    }
+
+    private static async Task<OutboxTransaction> BeginTransactionInternal(ISqlOutboxTransaction transaction, ContextBag context)
+    {
+        try
+        {
+            return await transaction.Begin(context).ConfigureAwait(false);
+        }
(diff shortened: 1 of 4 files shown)


src/SqlPersistence/Outbox/OutboxPersister.cs
```diff
@@ -33,7 +33,26 @@ public Task<OutboxTransaction> BeginTransaction(ContextBag context)
...
+            result = await transaction.Begin(context).ConfigureAwait(false);
+            return result;
+        }
+        catch (Exception e)
+        {
+            // A method that returns something that is disposable should not throw during the creation
+            // of the disposable resource. If it does the compiler generated code will not dispose anything
+            // therefore we need to dispose here to prevent the connection being returned to the pool being
+            // in a zombie state.
+            result?.Dispose();
```
SzymonPobiega, line 53: Shouldn't that be `transaction` and not `result?`? If we throw in `Begin` the result will be null and we won't dispose of the resources.

SzymonPobiega (changes requested):

tmasternak (approved):

SzymonPobiega (approved):

PR #565, opened 2020-11-04
Vendor tests grouped in namespaces to run setup fixture per vendor

ramonsmits:
Previously only one setup fixture existed that setup environment for only 2 databases. Now all vendor-specific tests are grouped in a product namespace containing a SetUpFixture for each. This results in the ability run product-specific tests without a dependency on setup fixture of a different vendor.

### DIFF
--- a/src/SqlPersistence.Tests/Outbox/MySqlOutboxPersisterTests.cs
+++ b/src/SqlPersistence.Tests/Outbox/MySqlOutboxPersisterTests.cs
@@ -1,23 +1,26 @@
-using System;
-using System.Data.Common;
-using NServiceBus.Persistence.Sql.ScriptBuilder;
-using NUnit.Framework;
-
-[TestFixture(false, false)]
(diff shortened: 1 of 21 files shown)


bording:
I'm not sure I follow the point of this PR. What do we gain by putting things into separate namespaces? 

Also, it seems like the setup methods for MySql and Oracle don't actually do anything.

ramonsmits:
@bording one time fixture setup is run per namespace. Previously, if you do not have for example Postgresql then all tests failed because the one time fixture setup would fail on psql not being available while plenty of tests like mssql would just run..

Setup methods for MySql and Oracle are doing "creating a connection". If that fails all tests in that namespace will fail if that wasn't possible.

bording:
> @bording one time fixture setup is run per namespace.

Aha! I didn't know that. Now I understand what this PR is doing.

bording (approved):

PR #587, opened 2020-12-29
Support SQL Server AlwaysEncrypted

lailabougria:
Support SQL Server Always Encrypted on Sql-persistence.

### DIFF
--- a/src/SqlPersistence/Outbox/OutboxPersister.cs
+++ b/src/SqlPersistence/Outbox/OutboxPersister.cs
@@ -82,8 +82,7 @@ class OutboxPersister : IOutboxStorage
                 command.AddParameter("MessageId", messageId);
 
                 // to avoid loading into memory SequentialAccess is required which means each fields needs to be accessed, but SequentialAccess is unsupported for SQL Server AlwaysEncrypted
-                var behavior = sqlDialect.GetBehavior(connection);
-                using (var dataReader = await command.ExecuteReaderAsync(behavior).ConfigureAwait(false))
+                using (var dataReader = await command.ExecuteReaderAsync(CommandBehavior.SequentialAccess | CommandBehavior.SingleRow).ConfigureAwait(false))
                 {
                     if (!await dataReader.ReadAsync().ConfigureAwait(false))
                     {
--- a/src/SqlPersistence/Outbox/OutboxPersister.cs
+++ b/src/SqlPersistence/Outbox/OutboxPersister.cs
@@ -17,17 +17,15 @@ class OutboxPersister : IOutboxStorage
     int cleanupBatchSize;
     OutboxCommands outboxCommands;
     Func<ISqlOutboxTransaction> outboxTransactionFactory;
-    bool isSequentialAccessSupported;
 
     public OutboxPersister(IConnectionManager connectionManager, SqlDialect sqlDialect, OutboxCommands outboxCommands,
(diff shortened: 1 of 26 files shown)


src/SqlPersistence/Outbox/OutboxPersister.cs
```diff
@@ -79,8 +81,14 @@ public async Task<OutboxMessage> Get(string messageId, ContextBag context)
                 command.CommandText = outboxCommands.Get;
                 command.Transaction = transaction;
                 command.AddParameter("MessageId", messageId);
-                // to avoid loading into memory SequentialAccess is required which means each fields needs to be accessed
-                using (var dataReader = await command.ExecuteReaderAsync(CommandBehavior.SingleRow | CommandBehavior.SequentialAccess).ConfigureAwait(false))
+
+                // to avoid loading into memory SequentialAccess is required which means each fields needs to be accessed, but SequentialAccess is unsupported for SQL Server AlwaysEncrypted
+                var behavior = CommandBehavior.SingleRow;
+                if (!connection.IsEncrypted())
```
SzymonPobiega, line 87: Can we store/cache this value somehow? I am afraid that parsing the connection string each time we execute a query will add significant cost (although not sure how to measure it).

Maybe the dialect could inspect if connection is encrypted first time someone calls `CreateCommand` and then set `IsEncrypted` property on the `CommandWrapper` it returns.

src/SqlPersistence/Outbox/OutboxPersister.cs
```diff
@@ -79,8 +81,10 @@ public async Task<OutboxMessage> Get(string messageId, ContextBag context)
                 command.CommandText = outboxCommands.Get;
                 command.Transaction = transaction;
                 command.AddParameter("MessageId", messageId);
-                // to avoid loading into memory SequentialAccess is required which means each fields needs to be accessed
-                using (var dataReader = await command.ExecuteReaderAsync(CommandBehavior.SingleRow | CommandBehavior.SequentialAccess).ConfigureAwait(false))
+
+                // to avoid loading into memory SequentialAccess is required which means each fields needs to be accessed, but SequentialAccess is unsupported for SQL Server AlwaysEncrypted
+                var behavior = sqlDialect.GetBehavior(connection);
```
SzymonPobiega, line 86: In line 79 the dialect is creating a command and then here we get the command behavior and apply it. I am wondering if we can't make it a bit simpler now that we have the `isEncrypted` in the dialect. I think we could change the `CommandWrapper` a bit:

```C#
public Task<DbDataReader> ExecuteReaderAsync(CommandBehavior behavior)
{
    var resultingBehavior = dialect.ModifyBehavior(behavior);
    return command.ExecuteReaderAsync(resultingBehavior);
}
```

and then change the `GetBehavior` into `ModifyBehavior`:

```C#
internal override CommandBehavior ModifyBehavior(CommandBehavior baseBehavior, DbConnection connection)
{
    if (!hasConnectionBeenInspectedForEncryption)
    {
        isConnectionEncrypted = connection.IsEncrypted();
        hasConnectionBeenInspectedForEncryption = true;
    }

    if (isConnectionEncrypted)
    {
        baseBehavior &= ~CommandBehavior.SequentialAccess; //Remove sequential access
    }

    return baseBehavior;
}
```
 
This would require passing the connection to the ctor of `CommandWrapper` but this should not be a big deal since it is available when the wrapper is created.

The benefit of this approach is that the code in the persisters would not need to be modified at all because it only passes the "base behaviors".

SzymonPobiega (approved):

PR #598, opened 2021-01-28
Dotnet tool for script generation

HEskandari:
- New project for cmdline tool to generate scripts
- Refactor to remove the special logging step (only used for the 'saga generation' step)


src/CommandLine/NServiceBus.Persistence.Sql.CommandLine.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <OutputType>Exe</OutputType>
+    <ToolCommandName>sql-persistence</ToolCommandName>
+    <PackAsTool>True</PackAsTool>
+    <Description>.NET Core global tool to generate scripts for NServiceBus Sql persistence</Description>
+    <GenerateDocumentationFile>false</GenerateDocumentationFile>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="McMaster.Extensions.CommandLineUtils" Version="3.1.0" />
+    <PackageReference Include="Particular.Packaging" Version="0.9.0" PrivateAssets="All" />
```
mikeminutillo, line 14: We might be getting this from the `Directory.Build.props` anyway, in which case we shouldn't have it here.

src/ScriptBuilderTask.Tests.Target/AssemblyInfo.cs
```diff
@@ -5,4 +5,4 @@
     MySqlScripts = true,
     PostgreSqlScripts = true,
     OracleScripts = true,
-    ScriptPromotionPath = "$(SolutionDir)$(ProjectDir)Postfix")]
\ No newline at end of file
+    ScriptPromotionPath = @"C:\Temp\Prod")]
```
mikeminutillo, line 8: I wonder if we should allow the user to specify arbitary variables and then have the msbuild script pass in these two. i.e. if we could say at the command line:

```
sql-persistence general assemblyname outputpath -v SolutionDir="Path\To\Solution" -v ProjectDir="Path\To\Project"
```

mikeminutillo (commented):
So far so good. I think the biggest thing is going to be deciding how and where this gets configured. Where precisely is the line between build-time configuration (msbuild) and runtime configuration (command line)? Can we override the attribute baked into the assembly if there's a good reason to do so? Will we deprecate the attribute and allow these settings to be controlled directly via .csproj properties?

src/ScriptBuilderTask.Tests.Target/AssemblyInfo.cs
HEskandari: Problem with this is that, if the tool is used outside the dev environment there may be no Solution/Project directory. It hinders the usage outside of the MSBuild integration.

Alternatively, the relative paths can be used, which means it'll be relative to the project directory (I think).

src/CommandLine/NServiceBus.Persistence.Sql.CommandLine.csproj
HEskandari: removed

src/ScriptBuilderTask.Tests.Target/AssemblyInfo.cs
HEskandari: Since we decided to keep the cmd tool separate to the MSBuild task, there'll be no need for this for now.

HEskandari:
@mikeminutillo had a bunch of commits to fix some broken / non-deterministic tests. I have added a couple of more options. Would be great if you could have another look.

src/ScriptBuilder/Writers/OutboxWriter.cs
```diff
@@ -1,21 +1,41 @@
...
+        }
+
+        CheckExistingFile(overwrite, dropPath);
         using (var writer = File.CreateText(dropPath))
         {
             OutboxScriptBuilder.BuildDropScript(writer, sqlDialect);
         }
     }
+
+    static void CheckExistingFile(bool overwrite, string sagaFileName)
```
mikeminutillo, line 34: `sagaFileName` should be `outboxFileName`. Maybe this should just be `fileName`?

src/ScriptBuilder/Writers/OutboxWriter.cs
```diff
@@ -1,21 +1,41 @@
...
 using NServiceBus.Persistence.Sql.ScriptBuilder;
 
 class OutboxWriter
 {
-    public static void WriteOutboxScript(string scriptPath, BuildSqlDialect sqlDialect)
+    public static void WriteOutboxScript(bool clean, bool overwrite, string scriptPath, BuildSqlDialect sqlDialect)
     {
         var createPath = Path.Combine(scriptPath, "Outbox_Create.sql");
-        File.Delete(createPath);
+        if (clean)
```
mikeminutillo, line 10: There's a lot of duplicated logic for each file and for each type of script we output. I wonder if it's worth pulling some of it into a base class and then having the outbox writer look something like this:

```csharp
class OutboxWriter : ScriptWriter
{
  ctor(clean, overwrite, scriptPath) : base(clean, overwrite, scriptPath) { ... }
  
  public void WriteOutboxScript(BuildSqlDialect sqlDialect)
  {
    WriteScript("Outbox_Create.sql", writer => OutboxScriptBuilder.BuildCreateScript(writer, sqlDialect));
    WriteScript("Outbox_Drop.sql", writer => OutboxScriptBuilder.BuildDropScript(writer, sqlDialect));
  }
}
```

src/ScriptBuilder/Writers/TimeoutWriter.cs
```diff
@@ -1,21 +1,41 @@
...
+        }
+
+        CheckExistingFile(overwrite, dropPath);
         using (var writer = File.CreateText(dropPath))
         {
             TimeoutScriptBuilder.BuildDropScript(writer, sqlDialect);
         }
     }
+
+    static void CheckExistingFile(bool overwrite, string sagaFileName)
```
mikeminutillo, line 34: Should be `timeoutFileName` or maybe just `fileName`?

src/ScriptBuilder/Writers/SubscriptionWriter.cs
```diff
@@ -1,21 +1,41 @@
...
+        }
+
+        CheckExistingFile(overwrite, dropPath);
         using (var writer = File.CreateText(dropPath))
         {
             SubscriptionScriptBuilder.BuildDropScript(writer, sqlDialect);
         }
     }
+
+    static void CheckExistingFile(bool overwrite, string sagaFileName)
```
mikeminutillo, line 34: Should be `subscriptionFileName` or maybe just `fileName`.

src/ScriptBuilder/Saga/AllSagaDefinitionReader.cs
```diff
@@ -27,12 +29,15 @@ public IEnumerable<SagaDefinition> GetSagas(Action<ErrorsException, TypeDefiniti
...
+                errors.Add(new Exception($"Error in '{type.FullName}' (Filename='{type.GetFileName()}'). Error: {exception.Message}", exception));
             }
         }
-        return sagas;
-    }
-
 
+        if (errors.Count > 0)
+        {
+            throw new AggregateException(errors);
```
mikeminutillo, line 38: The way we used to do this, we would still generate a partial set of scripts if there were errors. i.e. If you had 5 sagas and one of them failed, you would get 4 scripts _before_ the error was thrown.

src/ScriptBuilderTask.Tests.Target/AssemblyInfo.cs
```diff
@@ -5,4 +5,4 @@
     MySqlScripts = true,
     PostgreSqlScripts = true,
     OracleScripts = true,
-    ScriptPromotionPath = "$(SolutionDir)$(ProjectDir)Postfix")]
\ No newline at end of file
+    ScriptPromotionPath = @"$(SolutionDir)\Prod")]
```
mikeminutillo, line 8: Is this change still needed?

src/ScriptBuilderTask.Tests/InnerTaskTests.cs
```diff
@@ -12,11 +11,11 @@ public void IntegrationTest()
     {
         var testDirectory = TestContext.CurrentContext.TestDirectory;
         var temp = Path.Combine(testDirectory, "InnerTaskTemp");
-        if (!Directory.Exists(temp))
+        if (Directory.Exists(temp))
         {
-            return;
```
mikeminutillo, line 17: I'm confused by this change. Was this broken before? It sems like it was bailing out of the test early if the temp folder was missing, is that right?

src/ScriptBuilder/ScriptWriter.cs
```diff
@@ -6,52 +6,103 @@
...
+            this.promotionFinder = promotionFinder;
+        }
+
+        public static void Write(string assemblyPath, string targetDirectory, Func<string, string> promotionPathFinder)
+        {
+            var writer = new ScriptWriter(assemblyPath, targetDirectory, promotionFinder: promotionPathFinder);
+            writer.Generate();
+        }
+
+        string GetScriptPath() => Path.Combine(destinationDirectory, "NServiceBus.Persistence.Sql");
```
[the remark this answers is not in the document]
HEskandari: done

src/ScriptBuilder/Writers/OutboxWriter.cs
HEskandari: done

src/ScriptBuilder/Writers/TimeoutWriter.cs
HEskandari: won't be applicable after the refactoring to the base class.

src/ScriptBuilder/Writers/SubscriptionWriter.cs
HEskandari: won't be applicable after the refactoring to the base class.

src/ScriptBuilder/Saga/AllSagaDefinitionReader.cs
HEskandari: I reverted back this error logging changes, so when used in MSBuild it will have the same old behavior.

src/ScriptBuilderTask.Tests.Target/AssemblyInfo.cs
HEskandari: reverted back

src/ScriptBuilderTask.Tests/InnerTaskTests.cs
HEskandari: yes, this was incorrect. On the first, unless you create that directory manually the test would not generate anything.

src/ScriptBuilder/Writers/OutboxWriter.cs
HEskandari: won't be applicable after the refactoring to the base class.

mikeminutillo (approved):

PR #611, opened 2021-02-12
Upgraded to the SQL Server Transport with Core v8 transport seam

tmasternak:


src/SqlPersistence/SqlPersistence.csproj
```diff
@@ -10,7 +10,7 @@
 
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" Version="[12.0.1, 13.0.0)" />
-    <PackageReference Include="NServiceBus" Version="[8.0.0-alpha.644, 9.0.0)" />
+    <PackageReference Include="NServiceBus" Version="8.0.0-alpha.819" />
```
DavidBoike, line 13: This should stay a version range. It will need to be for packaging once we start releasing stuff.
```suggestion
    <PackageReference Include="NServiceBus" Version="[8.0.0-alpha.819, 9.0.0)" />
```

DavidBoike (approved):
Just the one thing.

PR #631, opened 2021-02-25
Add instructions to smoke test always-ecrypted

lailabougria:
We currently don't have automated tests running on an encrypted database.
Adding the instructions on how to smoke test this.


SzymonPobiega (approved):
Tested the procedure and it works :shipit:

PR #658, opened 2021-03-12
Sql Persistence: CancellationToken + Core Alpha 910

DavidBoike:


src/SqlPersistence/Subscription/SubscriptionPersister.cs
```diff
@@ -26,7 +25,7 @@ public SubscriptionPersister(IConnectionManager connectionManager, string tableP
         }
     }
 
-    public async Task Subscribe(Subscriber subscriber, MessageType messageType, ContextBag context)
+    public async Task Subscribe(Subscriber subscriber, MessageType messageType, ContextBag context, CancellationToken cancellationToken = default)
```
andreasohlund, line 28: Token needs to be passed on?

src/SqlPersistence/Subscription/SubscriptionPersister.cs
```diff
@@ -45,7 +44,7 @@ public async Task Subscribe(Subscriber subscriber, MessageType messageType, Cont
         ClearForMessageType(messageType);
     }
 
-    public async Task Unsubscribe(Subscriber subscriber, MessageType messageType, ContextBag context)
+    public async Task Unsubscribe(Subscriber subscriber, MessageType messageType, ContextBag context, CancellationToken cancellationToken = default)
```
andreasohlund, line 47: Pass the token on?

DavidBoike:
I think at this point it would be best to wait for the new analyzer before continuing with this PR. There's a lot of places for things to hide.

adamralph (approved):

adamralph:
@andreasohlund @DavidBoike this should be good to go now. I rebased to resolve conflicts and added 4 commits.

src/MsSqlMicrosoftDataClientSqlTransportAcceptanceTests/MsSqlMicrosoftDataClientSqlTransportAcceptanceTests.csproj
```diff
@@ -13,17 +13,17 @@
...
   </ItemGroup>
 
   <ItemGroup>
     <Compile Remove="$(PkgNServiceBus_AcceptanceTests_Sources)\**\*.cs" />
-    <Compile Include="$(PkgNServiceBus_AcceptanceTests_Sources)\contentFiles\cs\any\**\EndpointTemplates\*.cs" />
-    <Compile Include="$(PkgNServiceBus_AcceptanceTests_Sources)\contentFiles\cs\any\**\ScenarioDescriptors\*.cs" />
-    <Compile Include="$(PkgNServiceBus_AcceptanceTests_Sources)\contentFiles\cs\any\**\NServiceBusAcceptanceTest.cs" />
+    <Compile Include="$(PkgNServiceBus_AcceptanceTests_Sources)\contentFiles\cs\$(TargetFramework)\**\EndpointTemplates\*.cs" />
+    <Compile Include="$(PkgNServiceBus_AcceptanceTests_Sources)\contentFiles\cs\$(TargetFramework)\**\ScenarioDescriptors\*.cs" />
+    <Compile Include="$(PkgNServiceBus_AcceptanceTests_Sources)\contentFiles\cs\$(TargetFramework)\**\NServiceBusAcceptanceTest.cs" />
```
adamralph, line 26: Anyone know why this changed, and if it's valid?

DavidBoike: It's because:

1. The AT Sources package is monolithic and should be split into multiple packages. This junk is to get just those 3 support files but none of the regular acceptance tests themselves.
2. Particular.Packaging had to change how it bundles the assets in a NuGet package in order to support working correctly with the new mix of target frameworks we're supporting, not working if you try to do netstandard2.0 with net461 or something like that, so the package itself no longer contains an "any" directory, so this is just fixing the paths to find the right files in the right places.

DavidBoike:
@adamralph Your changes LGTM - I can't approve a PR I raised though.

PR #753, opened 2021-06-23
Fix build

mikeminutillo:
`IDE0004 Cast is redundant` is being applied by mistake to a conditional operator expression for language versions prior to 9.

Additionally, enabled Oracle tests on .NET Core for Windows.

### DIFF
--- a/src/SqlPersistence.PersistenceTests/PersistenceTestsConfiguration.cs
+++ b/src/SqlPersistence.PersistenceTests/PersistenceTestsConfiguration.cs
@@ -165,7 +165,7 @@
             ISqlOutboxTransaction transactionFactory()
             {
                 return transactionScopeMode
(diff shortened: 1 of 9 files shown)


mikeminutillo:
Force pushed to clean up history

mauroservienti:
For my understanding, why do those Oracle tests run only on Windows, despite being .NET Core tests?

tmasternak (approved):

mikeminutillo:
> For my understanding, why do those Oracle tests run only on Windows, despite being .NET Core tests?

@mauroservienti the linux build agents don't have access to an Oracle database to execute the tests against. When the tests were originally written there was not a .NET Core client driver for Oracle so the Oracles tests were completely excluded. Now they run on Windows at least. The next step would be to get Oracle running on Linux and then remove the `OnlyOnWindowsAttribute`.

mauroservienti (approved):
Tnanks @mikeminutillo

SzymonPobiega (approved):

PR #861, opened 2022-01-31
fix script runner obsoletions

HEskandari:


mikeminutillo (changes requested):
The compiler will not be able to distinguish between these overloads if none of the optional parameters have been provided. We might be better off dropping the obsoletions in this case and leaving the original API intact. Users that are not explicitly calling the parameter will not notice the difference. Users that are explicitly calling it are probably setting to false so the behavior in v8 matches their expectations without the parameter. Removing the parameter in v8 without prior obsoletion is not ideal but there will be plenty of other signifiers why it has happened.

HEskandari:
Closing this in favor of https://github.com/Particular/NServiceBus.Persistence.Sql/pull/862

PR #915, opened 2022-04-12
Adjusting to new synchronized storage seam

danielmarbach:


src/SqlPersistence/SynchronizedStorage/SqlStorageSessionFeature.cs
```diff
@@ -12,13 +12,9 @@ protected override void Setup(FeatureConfigurationContext context)
...
 
-        var sessionHolder = new CurrentSessionHolder();
-
         //Info cache can be null if Outbox is enabled but Sagas are disabled.
-        services.AddSingleton<ISynchronizedStorage>(provider => new SynchronizedStorage(connectionManager, provider.GetService<SagaInfoCache>(), sessionHolder));
-        services.AddSingleton<ISynchronizedStorageAdapter>(provider => new StorageAdapter(connectionManager, provider.GetService<SagaInfoCache>(), sqlDialect, sessionHolder));
-
-        services.AddTransient(_ => sessionHolder.Current);
-        context.Pipeline.Register(new CurrentSessionBehavior(sessionHolder), "Manages the lifecycle of the current session holder.");
+        // TODO: The connection manager design needs a revisit to avoid the closure
```
SzymonPobiega, line 16: Is this done or what? :grumpy-face:

src/SqlPersistence/SynchronizedStorage/SqlDialect_MsSqlServer.cs
```diff
@@ -49,13 +48,12 @@ public partial class MsSqlServer
...
                     // Other modes handled by creating a new session.
-                    return null;
+                    return (WasAdapted: false, Connection: null, Transaction: null, OwnsTransaction: false);
                 }
 
                 var connection = await connectionManager.OpenConnection(context.GetIncomingMessage(), cancellationToken).ConfigureAwait(false);
                 connection.EnlistTransaction(ambientTransaction);
-
-                return storageSessionFactory(connection, null, true);
+                return (WasAdapted: true, Connection: connection, Transaction: null, OwnsTransaction: false);
```
SzymonPobiega, line 56:
```suggestion
                return (WasAdapted: true, Connection: connection, Transaction: null, OwnsTransaction: true);
```

Because we create a connection here and the `StorageSession` needs to Dispose it and it only calls `Dispose` if `OwnsTransaction` is `true`. Because the `Transaction` part of the tuple is `null`, it will not commit the transaction and that is fine because in this case the (ambient) transaction is managed by the transport.

SzymonPobiega (approved):
Laila also approves :)

src/SqlPersistence/SynchronizedStorage/SqlStorageSessionFeature.cs
danielmarbach: See https://github.com/Particular/NServiceBus.Persistence.Sql/pull/915/commits/79411ffb2679789abdfab7b07a11a912725e0801

PR #929, opened 2022-05-02
Align feature dependency with other persisters

danielmarbach:
This PR aligns the feature dependencies with other persisters and moves the saga/outbox persistence check to a feature that depends on the SqlOutbox and the SqlSaga features


danielmarbach:
The check in [core doesn't work unfortunately](https://github.com/Particular/NServiceBus/issues/6378)

lailabougria (approved):

PR #938, opened 2022-05-12
Upgrade targets to Net6.0

tmasternak:
Removes netcore3.1 and adds net6.0 to component targets


tmasternak:
@DavidBoike @bording Pinging you as requested, but we'll move ahead and merge this as we're otherwise blocked to continue with our work in the TF.

The update of the NSB Core package will be done in another PR as that PR then contains the changes that are required to match the new persistence seam.

lailabougria (approved):

PR #949, opened 2022-06-15
Testing dotnet7-preview on master

awright18:
Testing dotnet7-preview on master

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,9 @@ jobs:
       - name: Setup .NET SDK
         uses: actions/setup-dotnet@v2
         with:
-          dotnet-version: 6.0.x
+          dotnet-version: 6.0.x |
+            7.0.x
+          include-prerelease:true
       - name: Build
         run: dotnet build src --configuration Release
       - name: Upload packages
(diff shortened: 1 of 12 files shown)


.github/workflows/ci.yml
```diff
@@ -45,7 +45,9 @@ jobs:
       - name: Setup .NET SDK
         uses: actions/setup-dotnet@v2
         with:
-          dotnet-version: 6.0.x
+          dotnet-version: 6.0.x |
+            7.0.x
```
DavidBoike, line 49: The CI run is not starting because the `|` does not mean "there is more on the next line", it means "I intend to have an indented multi-line thing below this.
```suggestion
          dotnet-version: |
            7.0.x
            6.0.x
```

Though I am curious what would have happened if 6 had been specified before 7. Which SDK would be used to run `dotnet build`? We usually do highest first, assuming that the first one will be "primary". I'll test this on a repo that doesn't require multiple cumulative hours of CI time.

.github/workflows/ci.yml
```diff
@@ -45,7 +45,9 @@ jobs:
...
+          include-prerelease:true
```
DavidBoike, line 51: Apparently this is a syntax error not to have a space too
```suggestion
          include-prerelease: true
```

PR #990, opened 2022-08-11
Adding dependancy to Test project for Dependabot

awright18:
Adding dependencies to the Test project so Dependabot will keep track of them and notify use of issues


DavidBoike (approved):
Approved except for fixing the analyzer diagnostics from the new SDK version.

PR #1015, opened 2022-09-12
Update release to use Windows for building

awright18:
Using Windows will allow the IL Merge to work on when building ScriptBuilderTask

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,7 +8,7 @@ env:
   DOTNET_NOLOGO: true
 jobs:
   release:
-    runs-on: ubuntu-20.04
+    runs-on: windows-2019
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -8,7 +8,7 @@ env:
   DOTNET_NOLOGO: true
 jobs:
   release:
-    runs-on: ubuntu-20.04
+    runs-on: windows-2019
```
DavidBoike, line 11: Need some comment so we remember later why we're deviating from the "standard"
```suggestion
    runs-on: windows-2019 # Windows required for ILMerge to work in ScriptBuilderTask
```

DavidBoike (approved):

PR #1017, opened 2022-09-12
Remove the msbuild path override and execute script manually

danielmarbach:
This approach should also get rid of the problems we have with referencing the msbuild task.


danielmarbach:
For the RC branch https://github.com/Particular/NServiceBus.Persistence.Sql/pull/1018

lailabougria (approved):

tmasternak (approved):

PR #1128, opened 2022-12-30
Correctly clean all dialects if none selected

### LINKED ISSUE #848: SqlPersistenceSettings properties are not respected during script generation
It's possible, and recommended in the docs, to configure your script generation using something like the following:

```
[assembly: SqlPersistenceSettings(
   MsSqlServerScripts = false,
   MySqlScripts = false,
   OracleScripts = true,
   PostgreSqlScripts = false,
   ProduceOutboxScripts = false,
   ProduceSubscriptionScripts = false,
   ProduceTimeoutScripts = false
   ProduceSagaScripts = true,
   ScriptPromotionPath = "SqlPromotionScripts")]
```
The documented expectation from the above config is that only saga scripts for Oracle will be generated. When run (v6.2.1 of SQL-P), only Oracle scripts are created (correct) but all outbox, subscription, saga and timeout scripts are created within that. The expectation is that only saga scripts would be created. 

A quick look at the code shows two things:
1) there are no tests verifying that these configs are working
2) the `Produce*` settings are defaulted to `true` while the database dialect settings are not.

mikeminutillo:
- Fixes #848 

When we generate scripts, we clean out the folder to remove any scripts generated from a previous run. We limit this to only specified dialects. If you don't specify a dialect (which is the case for MSBuild) it _should_ clean out all of the dialects. Note that this is a different selection of dialects than the ones exposed via the assembly attribute.

Unfortunately, this is was broken in https://github.com/Particular/NServiceBus.Persistence.Sql/commit/04999bcd14759a003d1071361477000fec6abbcd

This change fixes the cleanup so we go back to the prior behavior. If no dialect is specified, then all dialects are cleaned up.

## Backporting

- 7.x (This PR)
- 6.6.x ...

### DIFF
--- a/src/ScriptBuilder/ScriptGenerator.cs
+++ b/src/ScriptBuilder/ScriptGenerator.cs
@@ -129,7 +129,7 @@ public class ScriptGenerator
 
     IEnumerable<string> GetSelectedDialects()
     {
-        return dialectOptions.Any() ? dialectOptions.Select(d => d.ToString()) : Enum.GetNames(typeof(BuildSqlDialect));
+        return dialectOptions.Count == 0 ? Enum.GetNames(typeof(BuildSqlDialect)) : dialectOptions.Select(d => d.ToString());
     }
 
     static IEnumerable<string> GetKnownScripts(string dialectDirectory)
--- a/src/ScriptBuilder/ScriptGenerator.cs
+++ b/src/ScriptBuilder/ScriptGenerator.cs
@@ -129,7 +129,7 @@ public class ScriptGenerator
 
     IEnumerable<string> GetSelectedDialects()
     {
-        return dialectOptions != null ? dialectOptions.Select(d => d.ToString()) : Enum.GetNames(typeof(BuildSqlDialect));
+        return dialectOptions.Any() ? dialectOptions.Select(d => d.ToString()) : Enum.GetNames(typeof(BuildSqlDialect));
     }
 
     static IEnumerable<string> GetKnownScripts(string dialectDirectory)


WilliamBZA (approved):

src/ScriptBuilder/ScriptGenerator.cs
```diff
@@ -129,7 +129,7 @@ void PurgeDialectDirs(string scriptPath)
 
     IEnumerable<string> GetSelectedDialects()
     {
-        return dialectOptions != null ? dialectOptions.Select(d => d.ToString()) : Enum.GetNames(typeof(BuildSqlDialect));
+        return dialectOptions.Any() ? dialectOptions.Select(d => d.ToString()) : Enum.GetNames(typeof(BuildSqlDialect));
```
WilliamBZA, line 132: `dialectOptions` won't ever be null because it is initialized to an empty list in the ctor if it is passed in, so :+1:

danielmarbach: Given `dialectOptions` is a readonly collection I would prefer checking the count like it is done here https://github.com/Particular/NServiceBus.Persistence.Sql/blob/66f5ddec00ec000a444ac6e9f9287958df6ea7f1/src/ScriptBuilder/ScriptGenerator.cs#L98

danielmarbach (approved):

src/ScriptBuilder/ScriptGenerator.cs
mikeminutillo: Updated. @danielmarbach do you want to quickly look again

danielmarbach (approved):

src/ScriptBuilder/ScriptGenerator.cs
WilliamBZA: For those of us in the back, why is `.Count` better than `.Any()`?

PR #1167, opened 2023-03-16
[Spike] Additional outbox+saga persistence tests

timbussmann:
Testing whether SQL would be able to run these tests before considering adding them to the core persistence tests


timbussmann:
closing this as the test will be added via https://github.com/Particular/NServiceBus/pull/6706

PR #1249, opened 2023-07-27
Optimize TSQL outbox cleanup by reducing batch size and adding rowlock hint

ramonsmits:
Optimize TSQL outbox cleanup by

1. Reducing batch size to prevent lock escalation to page/table locks 
2. Add hint to prevent escalation from row to page locks to not conflict with regular incoming message outbox operations as DELETE does not have to be efficient and should avoid any lock escalation over performance

To prevent lock escalation to page/table locks the number of rows must not be over 4.000 or 5.000 rows

Google states multiple resources that either going over 4.000 or 5.000 rows results into lock escalation from row to page/table locks:

- https://dba.stackexchange.com/questions/284675/delete-rows-older-than-x-days-without-locking-table
- https://www.sqlservercentral.com/forums/topic/row-locks-not-escalating-to-table-locks-after-5000
- https://www.sqlpassion.at/archive/2016/05/09/lock-escalations-do-they-always-happen/

### DIFF
--- a/src/SqlPersistence/Outbox/OutboxPersister.cs
+++ b/src/SqlPersistence/Outbox/OutboxPersister.cs
@@ -18,7 +18,7 @@ class OutboxPersister : IOutboxStorage
     readonly Func<ISqlOutboxTransaction> outboxTransactionFactory;
 
     public OutboxPersister(IConnectionManager connectionManager, SqlDialect sqlDialect, OutboxCommands outboxCommands,
-        Func<ISqlOutboxTransaction> outboxTransactionFactory) 
+        Func<ISqlOutboxTransaction> outboxTransactionFactory)
     {
         this.connectionManager = connectionManager;
         this.sqlDialect = sqlDialect;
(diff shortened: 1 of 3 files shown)


andreasohlund (approved):

src/SqlPersistence/Outbox/OutboxPersister.cs
```diff
@@ -20,7 +20,7 @@ class OutboxPersister : IOutboxStorage
 
     public OutboxPersister(IConnectionManager connectionManager, SqlDialect sqlDialect, OutboxCommands outboxCommands,
         Func<ISqlOutboxTransaction> outboxTransactionFactory,
-        int cleanupBatchSize = 4000)
+        int cleanupBatchSize = 4000) // Keep below 4000 to prevent lock escalation
```
andreasohlund, line 23: Nothing seems to be using this options so perhaps remove it and hardcode?

andreasohlund (approved):

udidahan:
Given the approval, it seems like this was validated as having some benefit, so why was it not followed through on?

ramonsmits:
> Given the approval, it seems like this was validated as having some benefit, so why was it not followed through on?

No specific reason except this didn't show up in the github notifications and due to your comment it did.

udidahan:
I would have thought that the ping from the bot would have triggered the same notification, but ok.

PR #1254, opened 2023-07-31
Set Size to -1 to optimize SQL Server query plan execution

### LINKED ISSUE #873: Not specifying parameter size for commands force type conversion which results in higher database CPU usage
Not specifying the [size of the parameters](https://docs.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlparameter.size) forces the execution engine to convert parameters to the max size specified in the column definition. This increases CPU usage on the database server.

While this should apply to all commands issued by persister we have a user report specifically for the outbox inserts https://discuss.particular.net/t/nservicebus-inefficient-sql-with-persistence/2945

ramonsmits:
- Resolves #873

Size needs to be set to -1

Smoketested using the [SQL Outbox sample](https://docs.particular.net/samples/outbox/sql/):

- Cloned Core to Core v9 and target locally created nugets from this branch.
- Modified the sample to have a text property increase in length for each message send
- Ran the sample where a message and sent a bunch of messages
- Cleared the query cache `DBCC FREEPROCCACHE `
- Ran the following query to observe the query cache statistics:
	```sql
	SELECT
		databases.name,
		dm_exec_sql_text.text AS TSQL_Text,
		dm_exec_query_stats.execution_count
	FROM sys.dm_exec_query_stats 
	CROSS APPLY sys.dm_exec_sql_text(dm_exec_query_stats.plan_handle)
	INNER JOIN sys.databases
	ON dm_exec_sql_text.dbid = databases.database_id
	WHERE databases.name='NsbSamplesSqlOutbox'
	ORDER BY TSQL_Text
	```
- Observed in the result that it works:
	```sql
	(@MessageId nvarchar(4000),@Operations nvarchar(4000),@PersistenceVersion nvarchar(4000))  insert into [sender].[OutboxData]  (      MessageId,      Operations,      PersistenceVersion  )  values  (      @MessageId,      @Operations,      @PersistenceVersion  )
	(@MessageId nvarchar(4000),@Operations nvarchar(4000),@PersistenceVersion nvarchar(4000))  insert into [sender].[OutboxData]  (      MessageId,      Operations,      PersistenceVersion  )  values  (      @MessageId,      @Operations,      @PersistenceVersion  )
	(@MessageId nvarchar(4000),@Operations nvarchar(max) ,@PersistenceVersion nvarchar(4000))  insert into [receiver].[OutboxData]  (      MessageId,      Operations,      PersistenceVersion  )  values  (      @MessageId,      @Operations,      @PersistenceVersion  )
	(@MessageId nvarchar(4000),@Operations nvarchar(max) ,@PersistenceVersion nvarchar(4000))  insert into [receiver].[OutboxData]  (      MessageId,      Operations,      PersistenceVersion  )  values  (      @MessageId,      @Operations,      @PersistenceVersion  )
	```

### DIFF
--- a/src/SqlPersistence/Config/SqlDialect_MsSqlServer.cs
+++ b/src/SqlPersistence/Config/SqlDialect_MsSqlServer.cs
@@ -36,7 +36,7 @@ namespace NServiceBus
                     parameter.Value = charSegment.Array;
 
                     // Set to 4000 or -1 to improve query execution plan reuse
-                    // Must be when exceeding 4000 characters for nvarchar(max)  https://stackoverflow.com/a/973269/199551
+                    // Must be set when exceeding 4000 characters for nvarchar(max)  https://stackoverflow.com/a/973269/199551
                     parameter.Size = charSegment.Count > 4000 ? -1 : 4000;
                 }
                 else if (value is string stringValue)
@@ -44,7 +44,7 @@ namespace NServiceBus
                     parameter.Value = stringValue;
 
                     // Set to 4000 or -1 to improve query execution plan reuse
-                    // Must be when exceeding 4000 characters for nvarchar(max)  https://stackoverflow.com/a/973269/199551
+                    // Must be set when exceeding 4000 characters for nvarchar(max)  https://stackoverflow.com/a/973269/199551
                     parameter.Size = stringValue.Length > 4000 ? -1 : 4000;
                 }
                 else
--- a/src/SqlPersistence/Config/SqlDialect_MsSqlServer.cs
+++ b/src/SqlPersistence/Config/SqlDialect_MsSqlServer.cs
@@ -34,7 +34,18 @@ namespace NServiceBus
                 if (value is ArraySegment<char> charSegment)
                 {
                     parameter.Value = charSegment.Array;
-                    parameter.Size = charSegment.Count;
+
+                    // Set to 4000 or -1 to improve query execution plan reuse
+                    // Must be when exceeding 4000 characters for nvarchar(max)  https://stackoverflow.com/a/973269/199551
+                    parameter.Size = charSegment.Count > 4000 ? -1 : 4000;
+                }
+                else if (value is string stringValue)
+                {
+                    parameter.Value = stringValue;
+
+                    // Set to 4000 or -1 to improve query execution plan reuse
+                    // Must be when exceeding 4000 characters for nvarchar(max)  https://stackoverflow.com/a/973269/199551
+                    parameter.Size = stringValue.Length > 4000 ? -1 : 4000;
                 }
                 else
                 {
--- a/src/SqlPersistence/Config/SqlDialect_MsSqlServer.cs
+++ b/src/SqlPersistence/Config/SqlDialect_MsSqlServer.cs
@@ -26,29 +26,50 @@ namespace NServiceBus
 
             internal override void SetJsonParameterValue(DbParameter parameter, object value)
             {
-                SetParameterValue(parameter, value);
+                SetParameterValue(parameter, value, -1);
             }
 
-            internal override void SetParameterValue(DbParameter parameter, object value)
+            internal override void SetParameterValue(DbParameter parameter, object value, int lengthMax)
             {
+                const int defaultLength = 4000;
+
                 if (value is ArraySegment<char> charSegment)
                 {
+                    if (lengthMax == 0)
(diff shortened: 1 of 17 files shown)


mauroservienti:
@ramonsmits are you taking care of the smoke testing bit?

ramonsmits:
@mauroservienti done! As this is an enhancement what do you think? Only release this for the next major or should this be backported to the current release major?

mauroservienti:
I think it should be backported to the currently released major. That's the only release available to users, the next major is in the works but there is no ETA.

src/SqlPersistence/Config/SqlDialect_MsSqlServer.cs
```diff
@@ -34,7 +34,18 @@ internal override void SetParameterValue(DbParameter parameter, object value)
                 if (value is ArraySegment<char> charSegment)
                 {
                     parameter.Value = charSegment.Array;
-                    parameter.Size = charSegment.Count;
+
+                    // Set to 4000 or -1 to improve query execution plan reuse
+                    // Must be when exceeding 4000 characters for nvarchar(max)  https://stackoverflow.com/a/973269/199551
+                    parameter.Size = charSegment.Count > 4000 ? -1 : 4000;
```
tmasternak, line 40: We know what are the sizes of the columns in the tables the persisted is using. Would it be possible to change the signature of the method so that the callers would be forced to pass that information?

tmasternak (commented):
Nice improvement. I'm pretty sure Dapper is doing exactly that

ramonsmits:
Although these last commits use the actual lengths of the schema it feels error-prone. It would require additional refactoring to really make the lengths part of a central schema to get the length information which is a major code change.

Dapper uses `-1` as the default is a string unless the actually string length is less than 4000 and then uses 4000 UNLESS a explicit length was configured

- https://github.com/DapperLib/Dapper/blob/f0e61a740b0c2f5421a64aff0bdb4707ced1c46b/Dapper/DbString.cs#L21

I vote for reverting these changes revert to e0d21e0. That behavior is pretty much as dapper except that dapper sets length to -1 when its a null value. That logic doesn't make sense IMHO and we can keep the behavior at e0d21e0.

@tmasternak @mauroservienti what do you think?

mauroservienti:
Agreed. 👍 to revert to https://github.com/Particular/NServiceBus.Persistence.Sql/commit/e0d21e061de44c97a05ed07663a95fb1b401e5f4

src/SqlPersistence/Config/SqlDialect_MsSqlServer.cs
```diff
@@ -34,7 +34,18 @@ internal override void SetParameterValue(DbParameter parameter, object value)
                 if (value is ArraySegment<char> charSegment)
                 {
                     parameter.Value = charSegment.Array;
-                    parameter.Size = charSegment.Count;
+
+                    // Set to 4000 or -1 to improve query execution plan reuse
+                    // Must be when exceeding 4000 characters for nvarchar(max)  https://stackoverflow.com/a/973269/199551
```
mauroservienti, line 39: Some text is missing in the comment. Is this what you meant?

```suggestion
                    // Must be set when exceeding 4000 characters for nvarchar(max)  https://stackoverflow.com/a/973269/199551
```

src/SqlPersistence/Config/SqlDialect_MsSqlServer.cs
```diff
@@ -34,7 +34,18 @@ internal override void SetParameterValue(DbParameter parameter, object value)
...
+                    parameter.Size = charSegment.Count > 4000 ? -1 : 4000;
+                }
+                else if (value is string stringValue)
+                {
+                    parameter.Value = stringValue;
+
+                    // Set to 4000 or -1 to improve query execution plan reuse
+                    // Must be when exceeding 4000 characters for nvarchar(max)  https://stackoverflow.com/a/973269/199551
```
mauroservienti, line 47: Same as above

mauroservienti (approved):

src/SqlPersistence/Config/SqlDialect_MsSqlServer.cs
ramonsmits:
```suggestion
                    // Must be set when exceeding 4000 characters for nvarchar(max)  https://stackoverflow.com/a/973269/199551
```

PR #1331, opened 2023-11-14
Stored json gets corrupted on SqlServer after an update with a smaller sized payload

ramonsmits:
When updating saga state where the serialized json is smaller than whats already present in the database the updated payload is corrupted.

## Expected Behavior 

When string or array values are updated with shorted values that there will be not data left from previous versions:
```
First : `111111`
Second: `2222`
```
## Observed Behavior 

In version 7.0.3 data of the previous version will remain in the database column for SQL server.
```
First:  `111111`
Second: `222211` // Trailing `11` from previous value
```
## How to reproduce

1. Design a saga that has an array or string value
2. Create a saga instance that has a value with size X (i.e. `111`)
3. Update the saga value with a value that is less than X (i.e. `11`)
4. Query the SQL Server database table and observe that the value will have some trailing data from the previous value

## Cause

7.0.3 introduced logic to set the SQL command parameter length to either 4000 or -1 for strings and arrays. This to ensure that SQL server will not create a new query plan so it will use the query cache.

Unfortunately, this has a side effect that SQL Server does not remove trailing data in an existing column for INSERT/UPDATE if the new value is shorter than the previous.

## Resolution

The length should only be set to 4000/-1 for SELECT queries

#### Versions

`NServiceBus.Persistence.Sql` 7.0.3

### Additional Information

- Introduced by #1254 
- Only version 7.0.3 is affected when using SQL Server

### DIFF
--- a/src/SqlPersistence.PersistenceTests/When_updating_saga.cs
+++ b/src/SqlPersistence.PersistenceTests/When_updating_saga.cs
@@ -19,15 +19,15 @@
                 return; // Satisfy compiler
             }
 
-            var sagaData = new SagaState
+            var sagaData = new SagaWithCorrelationPropertyData
             {
-                CorrelationProperty = Guid.NewGuid().ToString(),
+                CorrelatedProperty = Guid.NewGuid().ToString(),
                 Payload = "very long state"
             };
 
             await SaveSaga(sagaData);
 
-            SagaState retrieved;
+            SagaWithCorrelationPropertyData retrieved;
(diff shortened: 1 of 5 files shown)


andreasohlund:
Seems like we need to use a separate table for the new test to not break others?

andreasohlund (approved):

SimonCropp:
what versions does this effect?

andreasohlund:
Only 7.0.3 (see "Who's affected")

PR #1483, opened 2024-05-22
Integrate with the PostgreSQL transport

SzymonPobiega:
Allow connection/transaction sharing between the PostgreSQL transport and the PostgreSQL dialect of the persistence in order to provide [atomic state change and message sending](https://docs.particular.net/nservicebus/handlers/accessing-data#message-de-duplication-strategies-local-transactions).


src/PostgreSqlTransportAcceptanceTests/ConfigureEndpointPostgreSqlTransport.cs
```diff
@@ -0,0 +1,67 @@
...
+        using (var conn = PostgreSqlConnectionBuilder.Build())
+        {
+            conn.Open();
+
+            //var queueAddresses = transport.ReceivingAddresses.ToList();
+            //foreach (var address in queueAddresses)
+            //{
+            //    TryDeleteTable(conn, address);
+            //    TryDeleteTable(conn, new QueueAddress(address.Table + ".Delayed", address.Schema, address.Catalog));
+            //}
```
tmasternak, line 37: I think, this can be removed - as the container with PostgreSQL DB is setup on each run

poornimanayar (approved):

dvdstelt (approved):

PR #1554, opened 2024-08-30
Acceptance test that verifies synchronized storage integration works even when no messages are sent

danielmarbach:
The PR also bumps to the transactional session beta1 and reverts it again to verify the behavior is still preserved.

Should be squashed.


danielmarbach:
Had to force push a comment update. Previously, both relevant commits were green

image

mauroservienti (approved):

PR #1645, opened 2024-12-06
Add test permutation for transacton scope mode

andreasohlund:


andreasohlund:
TransactionScope mode works so adding permutations to test properly. I must have had a bug in the test code previously that caused me to think that it was broken

SzymonPobiega (approved):

PR #1734, opened 2025-03-20
Set Transaction = null on commit

awright18:
Set Transaction = null on commit


ramonsmits:
Closing, as obsoleted by newer prs

PR #1497, opened 2024-06-19
Set to patch instead of minor

dvdstelt:

PR #1750, opened 2025-04-21
Set to patch instead of minor (#1497)

awright18:
Remove call to DisposeAsync from CompleteAsync


ramonsmits (changes requested):
Can't do this as this would cause the dispose to run sync as Core is doing a sync dispose.

Reason is listed in the comment :-)

PR #1870, opened 2025-12-04
Provide a way to disable SQL source generator via MSBuild flag

DavidBoike:
Unfortunately it's not possible in an incremental generator to turn off whole stages, and this generator is simple enough we've basically done all the work in the very first stage, so all we can really do is prevent the writing of the source file.

@bording please especially look at the first commit. I was copying from how Core includes its props file and I could easily have missed something.

### DIFF
--- a/src/AnalyzerTests/SagaMetadataGeneratorTests.cs
+++ b/src/AnalyzerTests/SagaMetadataGeneratorTests.cs
@@ -247,7 +247,7 @@ public class SagaMetadataGeneratorTests
 
         SourceGeneratorTest.ForIncrementalGenerator<SagaMetadataGenerator>()
             .WithSource(code)
-            .WithProperty("build_property.NServiceBusDisableSagaSourceGenerator", "true")
+            .WithProperty("build_property.NServiceBusSqlPersistenceDisableSagaSourceGenerator", "true")
             .AddReference(MetadataReference.CreateFromFile(typeof(SqlSagaAttribute).Assembly.Location))
             .WithGeneratorStages("SagaConfigureMethods", "SagaDetails", "Collected")
             .ShouldNotGenerateCode()
--- a/src/SqlPersistence/NServiceBus.Persistence.Sql.props
+++ b/src/SqlPersistence/NServiceBus.Persistence.Sql.props
@@ -1,7 +1,7 @@
 <Project>
 
   <ItemGroup Label="Compiler visible properties">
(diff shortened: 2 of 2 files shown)


src/NServiceBus.Persistence.Sql.Analyzer/SagaMetadataGenerator.cs
```diff
@@ -243,4 +249,6 @@ static void WriteProperty(StringBuilder b, string property, string? value, bool
 
     record SagaDetails(string SagaType, CorrelationDetails? CorrelationProperty, CorrelationDetails? TransitionalProperty, string? TableSuffix);
     readonly record struct CorrelationDetails(string Type, string Name);
+
+    const string DisableGeneratorPropertyName = "build_property.NServiceBusDisableSagaSourceGenerator";
```
danielmarbach:
```suggestion
    const string DisableGeneratorPropertyName = "build_property.NServiceBusSqlPersistenceDisableSagaSourceGenerator";
```

just in case we need to disable source gen in Core saga source generator too?

Which brings up a good point which nomenclature  should we be following?

MSBuild property names are XML element names, so:
	•	They must be valid XML element names
	•	No spaces
	•	No hyphens
	•	No leading digits
	•	Stick to letters, digits, and underscore, starting with a letter or underscore

Sou should we even embark on something like `NServiceBus_Component_Property`?

danielmarbach:
> Unfortunately it's not possible in an incremental generator to turn off whole stages, and this generator is simple enough we've basically done all the work in the very first stage, so all we can really do is prevent the writing of the source file.

I think we might be talking about two slightly different things.

Inside an incremental generator we can structure the pipeline so that it effectively does no work. For example, we can read an MSBuild property via AnalyzerConfigOptions and then filter everything out so no sources are produced and none of the heavier stages run. That part is fully supported.

What we can’t do from inside the generator is disable the generator itself. Once the project references the analyzer, Roslyn will still load it and call Initialize, even if we immediately return no values. If we want a true on/off switch where the generator is not loaded at all, that has to be controlled at the MSBuild level where the analyzer reference is added.

So the property can definitely be used to make the generator behave like a no-op, but it can’t prevent Roslyn from initializing it unless the reference is conditioned in MSBuild.

src/NServiceBus.Persistence.Sql.Analyzer/SagaMetadataGenerator.cs
danielmarbach: Yeah also not a big fan of underscores either.

PR #2043, opened 2026-07-14
Respect TableSuffix property of the SqlSaga attribute when generating the persistence manifest

### LINKED ISSUE #2042: SQL Persistence manifest reports wrong table names, ignoring TableSuffix, NameFilter and TablePrefix
### Symptoms

Incorrect table names are added to the manifest report. This does not affect runtime behavior.

### Who's affected

You are affected if you are using the generated manifest report for SQL persistence.

### Root cause

The persistence manifest built its table names from `manifest.Prefix` (the raw endpoint name) plus the saga's `EntityName`, while the runtime derives them from `settings.GetTablePrefix(...)` and the saga's table suffix. The two disagreed whenever any table-naming configuration was in play, so the manifest advertised tables that don't exist.

### Confirmed workarounds

No workaround exists, however this bug does not affect runtime behavior.

<details>
<summary>Original bug report</summary>
#### Description
When using Oracle persistence and a long saga class name (>27 chars) in combination with the SqlSaga attribute which contains a shorter name (<27 chars),an exception is logged on startup while creating the persistence manifest.

#### Expected behavior
No exception, the attribute is used.

#### Actual behavior
Exception logged:

> Saga 'ClassNameThatIsTooLongForOracleTooHandleSaga' contains more than 27 characters, which is not supported by SQL persistence using Oracle. Either disable Oracle script generation using the SqlPersistenceSettings assembly attribute, shorten the name of the saga, or specify an alternate table name by overriding the SqlSaga's TableSuffix property.

#### Versions
9.0.0

### Steps to reproduce

- Write a saga
```
[SqlSaga(tableSuffix: "ShortSaga")]
internal class ClassNameThatIsTooLongForOracleTooHandleSaga()
    : Saga<ClassNameThatIsTooLongForOracleTooHandleSagaData>
{
/* etc */
}
```

- Use Oracle persistence
- Run the application

### Relevant log output

```shell

```

#### Workarounds

No workaround is available, however saga read and write to the correct tables are not affect. Nothing is corrupted, no data is at risk. This is a diagnostics-only defect.

#### Additional information

**Outbox and subscription manifests also affected**

##### The manifest ignores `NameFilter`

There's a `SagaSettings.NameFilter(...)` hook that rewrites saga table names. The [runtime applies](https://github.com/Particular/NServiceBus.Persistence.Sql/blob/a9b0c8ed8664bde78b2aa6471e9c34fe47fea3bd/src/SqlPersistence/Saga/RuntimeSagaInfo.cs#L60) it, but the manifest doesn't.

#####  The manifest ignores `TablePrefix`

The real prefix comes from `settings.GetTablePrefix(endpointName)` which returns your custom `persistence.TablePrefix("Foo_")` if you set one. The manifest instead hardcodes `$"{manifest.Prefix}_"`, where `Prefix` is the **raw endpoint name**.
</details>

dvdvorle:
This ensures that the same name is used in the manifest as in other places. Prevents exceptions in case of long names when using Oracle.

Fixes #2042

### DIFF
--- a/src/SqlPersistence.Tests/Saga/SqlSagaFeatureTests.cs
+++ b/src/SqlPersistence.Tests/Saga/SqlSagaFeatureTests.cs
@@ -1,0 +1,47 @@
+using System.Linq;
+using System.Threading.Tasks;
+using NServiceBus;
+using NServiceBus.Persistence.Sql;
+using NServiceBus.Sagas;
+using NUnit.Framework;
+
+[TestFixture]
+public class SqlSagaFeatureTests
+{
+    [Test]
+    public void Saga_manifest_uses_sql_saga_table_suffix_for_oracle()
+    {
+        var metadata = SagaMetadata.CreateMany([typeof(SagaWithEntityNameLongerThanOracleAllows)]).Single();
+
+        var manifest = SqlSagaFeature.CreateSagaManifest(
+            metadata,
+            new SqlDialect.Oracle(),
+            "Endpoint_",
+            sagaName => sagaName);
+
+        Assert.That(manifest.TableName, Is.EqualTo("\"SHORTSAGATABLE\""));
+        Assert.That(manifest.Indexes.Single().Name, Is.EqualTo("Index_Correlation_CorrelationProperty"));
+        Assert.That(manifest.Indexes.Single().Columns, Is.EqualTo("CorrelationProperty"));
+    }
+
+    [SqlSaga(tableSuffix: "ShortSagaTable")]
+    public class SagaWithEntityNameLongerThanOracleAllows :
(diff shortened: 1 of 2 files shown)


AlexanderVanWoensel (approved):

warwickschroeder:
Hi @dvdvorle,

Thank you for your contribution! We’ve routed the PR to the appropriate team for evaluation.
 
If only small adjustments are needed, we’ll take care of them. If larger changes are required, we’ll either handle them ourselves (when we have capacity) or let you know what’s needed so you can choose whether to continue, or wait until we’re able to pick it up.
 
Kind regards,
The team at Particular

warwickschroeder:
@dvdvorle - Thank you for your contribution. Unfortunately we can't run our CI workflow on pull requests from forks, so I've opened a new pull request with your changes, along with some additional fixes: https://github.com/Particular/NServiceBus.Persistence.Sql/pull/2049

PR #2049, opened 2026-07-22
Use the configured table prefix, saga suffix, and name filter when generating the persistence manifest

### LINKED ISSUE #2042: SQL Persistence manifest reports wrong table names, ignoring TableSuffix, NameFilter and TablePrefix

warwickschroeder:
Closes: #2042
Built off PR: https://github.com/Particular/NServiceBus.Persistence.Sql/pull/2043

## Problem

The persistence manifest built its table names from `manifest.Prefix` (the raw endpoint name) plus the saga's `EntityName`, while the runtime derives them from `settings.GetTablePrefix(...)` and the saga's table suffix. The two disagreed whenever any table-naming configuration was in play, so the manifest advertised tables that don't exist.

The saga case also broke Oracle deployments. An entity name over Oracle's 27-character identifier limit is exactly what tableSuffix exists to work around, and the manifest ignored it.

## Change

Each of the three features now resolves manifest table names through the same code path the runtime uses:

- Sagas - `BuildSagaManifests` reads per-saga data via `SqlSagaTypeDataReader.GetTypeData` (the reader the runtime uses), applies the configured `NameFilter`, and builds the prefix with `GetTablePrefix`. This also fixes the correlation-property index name when `[SqlSaga]` overrides the mapper's choice.
- Outbox - endpoint-name resolution (including `ProcessorEndpointKey`) is extracted into `GetTablePrefix`, shared by `Setup` and the manifest so they can't drift.
- Subscriptions - uses `GetTablePrefix(settings.EndpointName())` instead of `manifest.Prefix`.

## Tests

8 unit tests across `SqlSagaFeatureTests` and `SqlOutboxFeatureTests`, covering the above.


src/SqlPersistence/Outbox/SqlOutboxFeature.cs
```diff
@@ -59,10 +59,7 @@ ISqlOutboxTransaction transactionFactory() => transactionScopeMode
 
         if (settings.TryGet<ManifestOutput.PersistenceManifest>(out var manifest))
         {
-            manifest.SetOutbox(() => new ManifestOutput.PersistenceManifest.OutboxManifest
-            {
-                TableName = sqlDialect.GetOutboxTableName($"{manifest.Prefix}_")
-            });
+            ConfigureOutboxManifest(manifest, settings, sqlDialect);
```
mikeminutillo: Given that we already retrieved the `tablePrefix` above, why not keep the original implementation here and pass it in rather than passing `settings` and retrieving it again? Or at least pass `tablePrefix` here and not `settings`?

src/SqlPersistence/Outbox/SqlOutboxFeature.cs
```diff
@@ -94,6 +91,20 @@ ISqlOutboxTransaction transactionFactory() => transactionScopeMode
             new OutboxCleaner(outboxPersister.RemoveEntriesOlderThan, b.GetRequiredService<CriticalError>().Raise, timeToKeepDeduplicationData, frequencyToRunCleanup, new AsyncTimer()));
     }
 
+    // The outbox may run against a processor endpoint rather than this endpoint, so the
+    // endpoint name is resolved here and shared by Setup and the manifest to stop the two drifting.
+    internal static string GetTablePrefix(IReadOnlySettings settings) =>
+        settings.GetTablePrefix(settings.GetOrDefault<string>(ProcessorEndpointKey) ?? settings.EndpointName());
+
+    internal static void ConfigureOutboxManifest(
```
mikeminutillo: I wonder if it is worth pulling all of these internal methods into a separate `ManifestBuilder` class or something so they are all together.

mikeminutillo (approved):

src/SqlPersistence/Outbox/SqlOutboxFeature.cs
warwickschroeder: It was mainly so a test could drive the whole chain, but you're right. Ill refactor.

abparticular (approved):

PR #2067, opened 2026-08-07
Remove 'tag' parameter from PostgreSQL setup

danielmarbach:
Removed the 'tag' parameter from the PostgreSQL setup step.


danielmarbach:
Good catch @tamararivera 👍

danielmarbach:
Replaced by https://github.com/Particular/NServiceBus.Persistence.Sql/pull/2068

